PR #2, opened 2016-05-24
add support for symbolic links in targz

jabgibson:
- symbolic links within a targz were not being written (failing instead). 
- Also, file mode was defaulting to 0666, rather than its original value.

When attempting to untar the Oracle Jdk1.8 tar file, it will fail, because there are 3 symbolic links in the tar file. This change adds support for symbolic links.

### DIFF
--- a/targz.go
+++ b/targz.go
@@ -128,7 +128,9 @@ func untarGzFile(tr *tar.Reader, header *tar.Header, destination string) error {
 	case tar.TypeDir:
 		return mkdir(filepath.Join(destination, header.Name))
 	case tar.TypeReg:
-		return writeNewFile(filepath.Join(destination, header.Name), tr)
+		return writeNewFile(filepath.Join(destination, header.Name), tr, header.FileInfo().Mode())
+	case tar.TypeSymlink:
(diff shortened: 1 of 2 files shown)


mholt:
:100: Awesome! Thanks for fixing that! We should add this case to the tests. Doing so should be easy enough, just add a symlink to the testdata folder. It could also be worth checking the permissions on each file to make sure they match the original. 

To help with that, I've just added CI tests to this repo. I'm not sure how that works for this PR, but I imagine syncing your branch with master would pull in the yml files that trigger the CI builds.

The first CI builds are about to complete, I hope they pass... hehe.

jabgibson:
I might retract this request, and do some more work. Untaring symbolic links is no issue, but taring them is failing, thus the test fails.

mholt:
If you'd like to leave the PR open and simply update your branch with a new commit to be able to tar symlinks, that'd be just fine.

jabgibson:
Looks like Windows is not liking something I did. I might have to boot up the old windows partition and check out what is going on. Has to do with the already compressed jpg file

PR #3, opened 2016-05-26
add support for symbolic links and preserve file mode.

jabgibson:
Currently, while decrompressing (extracting files), if any are a symbolic link, archiver will fail. This change adds in support for the tar(gz) format to open, and create tar archives with symbolic links.
Zip files, do not support symbolic links, but this update does allow them to exist, merely skipping over them. There appears to be an issue with zip files and symbolic links.

This change, also preserves the file mode, and chmod according to original file mode. (Not applicable on Windows build)
Working on testing this. Currently the included tests attempt to compress, decompress all the files in test data. I need to write a test that will prevent zip files from messing with the symbolic links.

### DIFF
--- a/zip.go
+++ b/zip.go
@@ -146,10 +146,8 @@ func writeNewFile(fpath string, in io.Reader, fm os.FileMode) error {
 	defer out.Close()
 
 	err = out.Chmod(fm)
-	if err != nil {
-		if runtime.GOOS != "windows" {
-			return fmt.Errorf("%s: changing file mode: %v", fpath, err)
-		}
+	if err != nil && runtime.GOOS != "windows" {
+		return fmt.Errorf("%s: changing file mode: %v", fpath, err)
 	}
 
 	_, err = io.Copy(out, in)
--- a/zip.go
+++ b/zip.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"runtime"
 	"strings"
 )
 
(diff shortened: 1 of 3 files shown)


jabgibson:
my git fu (or lack thereof) really jacked my last pull request. Working on a cleaner one. Apologies

zip.go
```diff
@@ -142,13 +145,34 @@ func writeNewFile(fpath string, in io.Reader) error {
 	}
 	defer out.Close()
 
+	err = out.Chmod(fm)
+	if err != nil {
+		if runtime.GOOS != "windows" {
```
mholt: `if err != nil && runtime.GOOS != "windows"`

mholt:
Cool - tests seem to be passing as-is, which is good. :+1: Might be worth adding a file mode check to the tests, I imagine that would only be an extra few lines of code to the test.

zip.go
jabgibson: Good find. thanks.

mholt:
@jabgibson I'm excited to merge this in to release a '1.0'. Any chance you'd have a few minutes to make the last couple changes (assert file mode in tests & consolidate that if statement)?

jabgibson:
@mholt done. Sorry about taking so long, working to too many projects at once.

mholt:
@jabgibson No need to apologize, I just can never tell if responses just get buried in email inboxes. But I definitely understand the busy aspect of the life of a programmer! :smile: You're a champ, thank you so much for contributing!

I'll give you push rights if you ever feel like contributing again, as a thank you.

PR #4, opened 2016-07-04
Update tarbz2.go to account for recent API changes

dsnet:
The github.com/dsnet/compress packages now require a ReaderConfig
or WriterConfig to be passed in to NewReader or NewWriter.
A nil Config is guaranteed to always pass.

### DIFF
--- a/tarbz2.go
+++ b/tarbz2.go
@@ -20,7 +20,10 @@ func TarBz2(tarbz2Path string, filePaths []string) error {
 	}
 	defer out.Close()
 
-	bz2Writer := bzip2.NewWriter(out)
+	bz2Writer, err := bzip2.NewWriter(out, nil)
+	if err != nil {
+		return fmt.Errorf("error compressing %s: %v", tarbz2Path, err)
+	}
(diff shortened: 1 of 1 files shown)


tarbz2.go
```diff
@@ -20,7 +20,7 @@ func TarBz2(tarbz2Path string, filePaths []string) error {
 	}
 	defer out.Close()
 
-	bz2Writer := bzip2.NewWriter(out)
+	bz2Writer, _ := bzip2.NewWriter(out, nil)
```
shurcooL: Why not actually check the error instead of ignoring it? Presumably it may have some non-nil value, perhaps in the future. If it's always going to be nil, `bzip2.NewWriter` wouldn't return it.

I suppose you're trying to make the least number of changes just to update the API and allow this package to compile, but... I can't help but mention this.

dsnet: Fair enough. Done.

shurcooL:
LGTM.

mholt:
Woo hoo! Thanks, this was generous of you. 😃

PR #6, opened 2016-07-08
can't `UntarGz` some archives

Songmu:
`tar.TypeRegA` should be cared. 

Following code doesn't work with Error `pt_linux_amd64/pt: unknown type flag:`.

```go
package main

import (
	"fmt"
	"io"
	"net/http"
	"os"
	"path"

	"github.com/mholt/archiver"
)

func main() {
	url := "https://github.com/monochromegane/the_platinum_searcher/releases/download/v2.1.2/pt_linux_amd64.tar.gz"
	resp, _ := http.Get(url)
	defer resp.Body.Close()
	fname := path.Base(url)
	f, _ := os.Create(fname)
	io.Copy(f, resp.Body)

	err := archiver.UntarGz(fname, "pt")
	if err != nil {
		fmt.Println(err.Error())
	}
}
```

This patch resolve it.

### DIFF
--- a/targz.go
+++ b/targz.go
@@ -128,7 +128,7 @@ func untarGzFile(tr *tar.Reader, header *tar.Header, destination string) error {
 	switch header.Typeflag {
 	case tar.TypeDir:
 		return mkdir(filepath.Join(destination, header.Name))
-	case tar.TypeReg:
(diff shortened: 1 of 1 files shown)


mholt:
Oh, weird, I never noticed that. :smile: Thanks!

PR #7, opened 2016-07-08
parsing header.Name (filepath) for subfolders

producerism:
While trying to unrar an archive with subfolders, the call to header.IsDir evaluates as false because the folder structure does not exist yet. This is a less than ideal way to parse the path for subfolders and create them as needed before header.IsDir is checked against.

### DIFF
--- a/rar.go
+++ b/rar.go
@@ -5,7 +5,6 @@ import (
 	"io"
 	"os"
 	"path/filepath"
-	"strings"
 
 	"github.com/nwaples/rardecode"
 )
@@ -31,21 +30,12 @@ func Unrar(source, destination string) error {
 		return fmt.Errorf("%s: failed to create reader: %v", source, err)
 	}
 
-	// we should only parse for subfolders once per unrar request
-	subfoldersCreated := false
-
 	for {
 		header, err := rr.Next()
 		if err == io.EOF {
 			break
 		} else if err != nil {
 			return err
-		}
-
-		// make subfolders for this path
-		if !subfoldersCreated {
-			makeSubfolders(filepath.Dir(header.Name), destination)
-			subfoldersCreated = true
 		}
 
 		if header.IsDir {
@@ -56,6 +46,11 @@ func Unrar(source, destination string) error {
 			continue
 		}
 
+		err = mkdir(filepath.Dir(filepath.Join(destination, header.Name)))
+		if err != nil {
+			return err
+		}
+
 		err = writeNewFile(filepath.Join(destination, header.Name), rr, header.Mode())
 		if err != nil {
 			return err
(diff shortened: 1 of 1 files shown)


rar.go
```diff
@@ -37,6 +38,23 @@ func Unrar(source, destination string) error {
 		} else if err != nil {
 			return err
 		}
+		
+		pathComponents := strings.Split(header.Name, "/")
```
mholt: You could do `filepath.Dir(header.Name)` and avoid that `if pi == len(pathComponents) - 1` check below.

mholt: So is "/" always used in rar files, even on Windows?

mholt:
Thanks for finding a fix! You beat me to it. :smile:

rar.go
```diff
@@ -37,6 +38,23 @@ func Unrar(source, destination string) error {
...
+			// the last component of the path will be the file
+			// so ignore it, since we only want to create folders
+			if pi == len(pathComponents)-1 {
+				continue
+			}
+
+			// check to see if the path exists already
+			if stat, err := os.Stat(destination + path); err != nil || !stat.IsDir() {
+				// make the directory
+				mkdir(destination + path)
```
mholt: This function will make all parent folders necessary to make the folder at `destination+path` so it should only need be called once.

rar.go
producerism: I have been testing on windows exclusively so far, so I can confirm that the forward-slash "/" is used on windows. Haven't tested on other platforms yet.

rar.go
producerism: The issue is that header only gets set in the for loop, and this needs to be called after header is set, but before header.isDir is evaluated.

producerism:
Made another update that prevents subfolder creation from being called multiple times. Still not 100% ideal solution, since user might want to unrar the same file twice, and folders may have been deleted in between calls to unrar. Open to suggestions on better implementation.

producerism:
Updated again with a simpler (better) way of making sure createSubfolders() is only called once per call to Unrar()

mholt:
Hey, nice improvements!

I think I was able to fix it simply with this just before the `writeNewFile` line:

```go
err = mkdir(filepath.Dir(filepath.Join(destination, header.Name)))
if err != nil {
	return err
}
```

Although I'm not 100% sure what the rar file you sent me is supposed to look like since I didn't ask for its zip equivalent. But I saw strange behavior before my change and it looks good now. Can you see if this works for you? If so I prefer it because it's much simpler.

producerism:
I can confirm that does work. Updated my branch accordingly.

mholt:
Cool, thanks for your help!

PR #8, opened 2016-08-08
fix copying contents: archive/tar: write too long

shenjh369:
When file is in use and size getting increased , create compressed file will failed and error is "copying contents: archive/tar: write too long".  Changing to use CopyN to copy only the file size at that time,  also checks io.EOF if file becomes small.

### DIFF
--- a/targz.go
+++ b/targz.go
@@ -95,7 +95,7 @@ func tarFile(tarWriter *tar.Writer, source, dest string) error {
 			}
 			defer file.Close()
 
-			_, err = io.Copy(tarWriter, file)
+			_, err = io.CopyN(tarWriter, file,info.Size())
 			if err != nil {
 				return fmt.Errorf("%s: copying contents: %v", path, err)
 			}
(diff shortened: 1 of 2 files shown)


mholt:
Interesting, thanks for the fix! Could you please run `go fmt` before I merge this in? Thanks.

mholt:
Thank you!

shenjh369:
Sorry about the format,  I will pay attention to it next time, : )

halfcrazy:
We reproduce this error in v4.0.0-alpha.8

https://github.com/mholt/archiver/blob/v4.0.0-alpha.8/archiver.go#L224

mholt:
@halfcrazy Are you sure? Can you please open a new issue with instructions to reproduce the bug?

halfcrazy:
> @halfcrazy Are you sure? Can you please open a new issue with instructions to reproduce the bug?

Yes, I open a new issue with reproduce instructions.

PR #9, opened 2016-08-20
use filepath.Dir() instead of path.Dir()

whatalnk:
Please use filepath.Dir(), not path.Dir().

In Windows, following code doesn't work with Error
 `peco_windows_amd64\peco_windows_amd64\Changes: creating new file: open peco_windows_amd64\peco_windows_amd64\Changes: The system cannot find the path specified.`

```
package main

import (
	"fmt"
	"io"
	"net/http"
	"os"
	"path"

	"github.com/mholt/archiver"
)

func main() {
	url := "https://github.com/peco/peco/releases/download/v0.4.0/peco_windows_amd64.zip"
	resp, _ := http.Get(url)
	defer resp.Body.Close()
	fname := path.Base(url)
	f, _ := os.Create(fname)
	io.Copy(f, resp.Body)

	err := archiver.Unzip(fname, "peco_windows_amd64")
	if err != nil {
		fmt.Println(err.Error())
	}
}
  
```

And following code doesn't extract .tar.gz file including symlink with Error: `dest\hoge.txt: creating new file: open dest\hoge.txt: The system cannot find the path specified.`

```
$ tar -tvf hoge.tar.gz
-rw-r--r-- username/197121     0 2016-08-20 14:35 hoge.txt
lrwxrwxrwx username/197121     0 2016-08-20 14:12 link -> hoge.txt
```

```
package main

import (
	"fmt"

	"github.com/mholt/archiver"
)

func main() {
	err := archiver.UntarGz("hoge.tar.gz", "dest")
	if err != nil {
		fmt.Println(err.Error())
	}
```

### DIFF
--- a/zip.go
+++ b/zip.go
@@ -135,7 +135,7 @@ func unzipFile(zf *zip.File, destination string) error {
 }
 
 func writeNewFile(fpath string, in io.Reader, fm os.FileMode) error {
-	err := os.MkdirAll(path.Dir(fpath), 0755)
+	err := os.MkdirAll(filepath.Dir(fpath), 0755)
 	if err != nil {
 		return fmt.Errorf("%s: making directory for file: %v", fpath, err)
 	}
@@ -159,7 +159,7 @@ func writeNewFile(fpath string, in io.Reader, fm os.FileMode) error {
 }
 
 func writeNewSymbolicLink(fpath string, target string) error {
-	err := os.MkdirAll(path.Dir(fpath), 0755)
+	err := os.MkdirAll(filepath.Dir(fpath), 0755)
 	if err != nil {
 		return fmt.Errorf("%s: making directory for file: %v", fpath, err)
 	}


shurcooL:
LGTM.

I think you should also look at line 72. It uses `path.Ext`, but should be using [`filepath.Ext`](https://godoc.org/path/filepath#Ext) since it's dealing with filepaths, not /-separated paths.

That way the entire `path` package can be removed, which is a win (given the code in this file deals primarily with files).

mholt:
Hmm, so are these two separate issues? Should we just fix both with this PR?

shurcooL:
They're both incorrect uses of path vs filepath package, but on different lines. The two addressed in this PR so far are more critical than the one I pointed out. They can either be fixed in the same PR or I'm different PRs. Completely up to you! :)

mholt:
I'll accept this PR either way - too bad the tests don't reflect this issue... would it be difficult to add a case that catches this?

shurcooL:
> too bad the tests don't reflect this issue... would it be difficult to add a case that catches this?

Do you already have tests that cover the relevant funcs? If so, you'd just need to run those tests on Windows (i.e., GOOS=windows).

mholt:
The CI tests do run on Windows. It's just that all I've implemented so far is a symmetric test, where it compresses then decompresses the contents. If the same bytes come out as went in, the test passes. I guess I could add unit tests, but I'm wondering why this error wasn't caught with the existing tests, since we do use folders in the test data...

shurcooL:
I see. That's a good question then.

In any case, it's possible (and advisable) to verify correctness of this patch by human verification. `os.MkdirAll` deals with system paths, so `filepath.Dir` should be used, not `path.Dir`.

whatalnk:
First, create a .zip file using archiver.

```
package main

import (
	"fmt"

	"github.com/mholt/archiver"
)

func main() {
	err := archiver.Zip("hoge.zip", []string{"hoge.txt", "folder"})

	if err != nil {
		fmt.Println(err.Error())
	}
}

```

Then, extract it.

```
package main

import (
	"fmt"

	"github.com/mholt/archiver"
)

func main() {
	err := archiver.Unzip("hoge.zip", "dest")

	if err != nil {
		fmt.Println(err.Error())
	}
}
```

* When `dest` already exists, `Unzip()` work fine.
* When `dest` does not exist, `Unzip()` does not work with error:  `dest\hoge.txt: creating new file: open dest\hoge.txt: The system cannot find the path specified.`

However, if I create .zip from Windows menu, `Unzip()` does not work whether dest exist or not.

---

They have different `FileHeader` 

```
package main

import (
	"archive/zip"
	"fmt"
	"log"
	"os"
)

func main() {
	fn := os.Args[1]
	r, err := zip.OpenReader(fn)
	if err != nil {
		log.Fatal(err)
	}
	defer r.Close()

	for _, f := range r.File {
		fmt.Println(f.Name)
	}
}
```

* .zip file created by `Unzip()`
```
> go run main.go hoge.zip

hoge.txt    
folder/         
folder\text.txt  
```

* .zip file created by Windows menu
```
> go run main.go fromWinMenu.zip

folder/text.txt
hoge.txt
```

mholt:
^ Ah. Oops. So this PR fixes that problem, @whatalnk ? (Also @shurcooL I see what you mean, and will definitely be merging this PR. Just want to understand the implications fully.)

I haven't had a chance to check myself -- been busy with some things -- but I appreciate your help!

whatalnk:
This PR fixes the problem of `Unzip()` in windows. 

In Windows, 

* `path.Dir("folder\\file.txt")` is `.`

and, 

* `filepath.Join("destination", "file.txt")` is `destination\\file.txt`

So, `os.MkdirAll(path.Dir(fpath), 0755)` in [L138](https://github.com/mholt/archiver/blob/master/zip.go#L138) means `os.MkdirAll(".", 0755)`, because `fpath` is made by `filepath.Join()` [(L134)](https://github.com/mholt/archiver/blob/master/zip.go#L134). 

This PR fixes this problem. 
`filepath.Dir("folder\\file.txt")` is `folder`

However, this PR may be incomplete.

The reason why this error wasn't caught with the existing tests is parent directory is always created before `writeNewFile()`. 

In `test.zip` created by `Zip()`, parent directory is listed before the children files. 
```
testdata/
testdata\already-compressed.jpg
testdata\proverbs/
testdata\proverbs\extra/
testdata\proverbs\extra\proverb3.txt
testdata\proverbs\proverb1.txt
testdata\proverbs\proverb2.txt
testdata\quote1.txt
```
So, parent directory is created by `mkdir()` in `unzipFile()` in [L124-L126](https://github.com/mholt/archiver/blob/master/zip.go#L124-L126) before children files.

Following change in `zip_test.go` will produce error.

```
diff --git a/zip_test.go b/zip_test.go
index 3a1aa86..8f016ab 100644
--- a/zip_test.go
+++ b/zip_test.go
@@ -21,7 +21,7 @@ func symmetricTest(t *testing.T, ext string, cf CompressFunc, dcf DecompressFunc
 
 	// Test creating archive
 	outfile := filepath.Join(tmp, "test"+ext)
-	err = cf(outfile, []string{"testdata"})
+	err = cf(outfile, []string{".gitignore", "testdata"})
 	if err != nil {
 		t.Fatalf("making archive: didn't expect an error, but got: %v", err)
 	}
@@ -34,7 +34,7 @@ func symmetricTest(t *testing.T, ext string, cf CompressFunc, dcf DecompressFunc
 
 	// Test extracting archive
 	dest := filepath.Join(tmp, "extraction_test")
-	os.Mkdir(dest, 0755)
+	// os.Mkdir(dest, 0755)
 	err = dcf(outfile, dest)
 	if err != nil {
 		t.Fatalf("extracting archive: didn't expect an error, but got: %v", err)
```

mholt:
Thank you for the explanation @whatalnk! Makes sense to me now. Appreciate your contribution.

PR #11, opened 2016-10-08
refactor

tw4452852:
Just separate into three parts:

1. Introduce `Archiver` interface.

2. For better maintaining, Separate ` targz.go` into `tar.go` and `targz.go` and gather all kinds of tests into `archiver.go`. 

3. Setup `Open` and `Make` benchmark for further work.

Signed-off-by: Tw <tw19881113@gmail.com>


tw4452852:
Just find this interesting project, like it. So I did some refactor and want to improve its performance further.

BTW, I used subtest and subbenchmark introducing in 1.7

rar.go
```diff
@@ -5,20 +5,32 @@ import (
...
 
 	"github.com/nwaples/rardecode"
 )
 
-// Rar makes a .rar archive, but this is not implemented because
+func init() {
+	RegisterFormat("Rar", r{})
+}
+
+type r struct{}
```
mholt: Maybe a more descriptive type name? Like rarFormat or something.

tar.go
```diff
@@ -0,0 +1,154 @@
...
+)
+
+func init() {
+	RegisterFormat("Tar", t{})
+}
+
+type t struct{}
+
+func (t) Match(filename string) bool {
+	// TODO: read file header to identify the format
```
mholt: Note that when creating archives, the file won't exist yet, so we'd still have to fall back to the filename. (Not really relevant for this PR but just thought I'd mention it.)

tarbz2.go
```diff
@@ -4,16 +4,28 @@ import (
...
 
 	"github.com/dsnet/compress/bzip2"
 )
 
-// TarBz2 creates a .tar.bz2 file at tarbz2Path containing
+func init() {
+	RegisterFormat("TarBz2", tb{})
+}
+
+type tb struct{}
```
mholt: Again, more descriptive name, like tarBz2Format maybe?

targz.go
```diff
@@ -4,34 +4,26 @@ import (
...
-		return fmt.Errorf("error creating %s: %v", tarPath, err)
-	}
-	defer out.Close()
+func init() {
+	RegisterFormat("TarGz", tg{})
+}
 
-	tarWriter := tar.NewWriter(out)
-	defer tarWriter.Close()
+type tg struct{}
```
mholt: tarGzFormat

zip.go
```diff
@@ -9,19 +9,29 @@ import (
...
-	"runtime"
 	"strings"
 )
 
-// Zip creates a .zip file in the location zipPath containing
+func init() {
+	RegisterFormat("Zip", z{})
+}
+
+type z struct{}
```
mholt: zipFormat

archiver_test.go
```diff
@@ -8,23 +8,33 @@ import (
...
+				t.Skip("not supported")
+			}
+			symmetricTest(t, name, ar)
+		})
+	}
 }
 
 // symmetricTest performs a symmetric test by using mf to make an archive
 // from the test corpus, then using of to open the archive and comparing
 // the contents to ensure they are equal.
```
mholt: The godoc here needs slight updating so it refers to `ar` instead of `mf` and `of`.

tar.go
tw4452852: Yes. The original intention here is that we'd better identify the format according to the file header instead of the extension of file name which may be fake. For creating archives, maybe there is no need to use `Match` for checking (a crazy user may create a tarball named `crazy.zip` if he wants...).

mholt:
Cool, thank you!

PR #12, opened 2016-10-10
export all supported formats

tw4452852:
Signed-off-by: Tw <tw19881113@gmail.com>

### DIFF
--- a/rar.go
+++ b/rar.go
@@ -10,8 +10,11 @@ import (
 	"github.com/nwaples/rardecode"
 )
 
+// Rar is for RAR archive format
+var Rar rarFormat
+
 func init() {
-	RegisterFormat("Rar", rarFormat{})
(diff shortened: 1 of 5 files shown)


rar.go
```diff
@@ -10,8 +10,12 @@ import (
 	"github.com/nwaples/rardecode"
 )
 
+// Rar is for RAR archive format
+var Rar Archiver
```
mholt: Can you just initialize this here, instead of in init? `var Are = RarFormat` for this one, and similarly for the other formats.

tw4452852: Done.

tw4452852:
Hi @mholt , could we just remove go1.6 in CI, it's just noisy...

mholt:
Yes, that's fine. I'm mobile or I would just do it right now. Good idea.

PR #17, opened 2016-12-02
Add hard link support

myfreeweb:
This was necessary for extracting [the FreeBSD base system](http://ftp.freebsd.org/pub/FreeBSD/releases/amd64/11.0-RELEASE/base.txz).

### DIFF
--- a/tar.go
+++ b/tar.go
@@ -151,6 +151,8 @@ func untarFile(tr *tar.Reader, header *tar.Header, destination string) error {
 		return writeNewFile(filepath.Join(destination, header.Name), tr, header.FileInfo().Mode())
 	case tar.TypeSymlink:
 		return writeNewSymbolicLink(filepath.Join(destination, header.Name), header.Linkname)
+	case tar.TypeLink:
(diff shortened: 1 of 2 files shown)


tar.go
```diff
@@ -151,6 +151,8 @@ func untarFile(tr *tar.Reader, header *tar.Header, destination string) error {
 		return writeNewFile(filepath.Join(destination, header.Name), tr, header.FileInfo().Mode())
 	case tar.TypeSymlink:
 		return writeNewSymbolicLink(filepath.Join(destination, header.Name), header.Linkname)
+	case tar.TypeLink:
+		return writeNewHardLink(filepath.Join(destination, header.Name), filepath.Join(destination, header.Linkname))
```
mholt: I'm not super familiar with how this should work -- notice that the line above for symbolic links doesn't join the `destination` and `header.Linkname` values like you do here; is it supposed to be different?

myfreeweb: That's what I tried first, of course — didn't work. It needs full paths.

mholt:
Great, thanks for making this better! :+1: Really appreciate it.

PR #18, opened 2016-12-15
Identify archive format from file header

tatsushid:
This adds an archive format identification by checking a file's header, which is commented as Todo in `Match` methods in each format codes. But despite this, `Match` method still prefers a file extension identification to this and uses this only when a file doesn't have an expected extension so its behavior would not change so much.

There are a few more points should be noted

- Added methods only return `bool` value, without an error value not to change `Match` method interface
- Added methods don't care if a passed `filename` is actually accessible or not and if not, they silently return `false`

### DIFF
--- a/archiver_test.go
+++ b/archiver_test.go
@@ -39,6 +39,10 @@ func symmetricTest(t *testing.T, name string, ar Archiver) {
 		t.Fatalf("making archive: didn't expect an error, but got: %v", err)
 	}
 
+	if !ar.Match(outfile) {
+		t.Fatalf("identifying format should be 'true', but got 'false'")
+	}
+
 	var expectedFileCount int
 	filepath.Walk("testdata", func(fpath string, info os.FileInfo, err error) error {
 		expectedFileCount++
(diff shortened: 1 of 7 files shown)


archiver_test.go
```diff
@@ -39,6 +39,10 @@ func symmetricTest(t *testing.T, name string, ar Archiver) {
 		t.Fatalf("making archive: didn't expect an error, but got: %v", err)
 	}
 
+	if !ar.Match(outfile) {
+		t.Fatalf("identifying format should be 'true', but got 'false'")
+	}
```
mholt: It looks like this only checks the file extension and not the file header code, maybe we should add another few lines to the test to call those functions to verify they are working too.

tatsushid: At the first commit 183d279 of the patch series, I confirmed the test failed because the `outfile` value is e.g. `test-TarGz`, without an extension at testing .tar.gz. After adding checking file header implementations, the test is passed.

mholt: Oh good :)

PR #19, opened 2017-01-11
tar: Add TypeChar, TypeBlock and TypeFifo header flags support

nhlfr:
Character devices, block devices and named pipes have the separate types, but at the same time they can be extracted like the other files and they're handled well by archive/tar library. Supporting these headers is necessary for extracting Linux rootfs archives.


nhlfr:
I was testing this on https://uk.images.linuxcontainers.org/images/centos/7/amd64/default/20170110_02:16/rootfs.tar.xz

mholt:
Ah, great, thanks. Didn't even think about those. :+1:

PR #24, opened 2017-02-15
Added LZ4 compression

railsmechanic:


mholt:
Cool, thanks for your contribution!

archiver_test.go
```diff
@@ -74,7 +74,7 @@ func symmetricTest(t *testing.T, name string, ar Archiver) {
 
 		if info.IsDir() {
 			// stat dir instead of read file
-			_, err := os.Stat(origPath)
+			_, err = os.Stat(origPath)
```
shurcooL: Just a nitpicky question out of curiosity, what was the motivation/rationale behind this change?

railsmechanic: My IDE indicates that this `err` is shadowed by the definition at line 70. 😬

mholt: I noticed this, but it looks harmless in this case.

railsmechanic: Absolutely! A harmless change...

shurcooL: I see, thanks.

I would want to discuss whether a shadowed variable with name `err` and type `error` is something that warrants issuing warnings about, but that's a discussion to have elsewhere.

mholt: I think it depends largely on the context. That might be why go vet doesn't issue warnings about it...

railsmechanic: It's great being a software developer, where a single colon is a starting point for a conversation. 👍

PR #36, opened 2017-07-12
Mod:fix bug for symbolic link file while making tar

yiduyangyi:
# problem
while making a tar file, if a source file is a symbolic, then the link information is wrong, just link to the file's path.

# fix
check if a file is a symbolic link, if it is , get the link information , then write to the tar file info header

### DIFF
--- a/tar.go
+++ b/tar.go
@@ -145,6 +145,20 @@ func tarFile(tarWriter *tar.Writer, source, dest string) error {
 			header.Name += "/"
 		}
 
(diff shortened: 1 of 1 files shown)


mholt:
Thanks for your contribution! Could you please add a test (feel free to add a symlink and/or file to the testdata folder) to verify this works?

mholt:
Closing due to inactivity. :( Will like to see this merged, but again, it'll need tests and the CI checks have to pass.

Issue #21, opened 2017-01-31
The symbolic links are not preserved when unpacking a zip (other formats not tested)


Skywalker13:
See subject

oaleynik:
@Skywalker13 did you find solution for this?

Skywalker13:
Yes but I use archive/zip instead:
```go
package main

import (
	"archive/zip"
	"io"
	"os"
	"path/filepath"
)

// Unzip a file to a destination and preserve the symlinks
// Based on http://stackoverflow.com/a/24792688/842097 with symlink additions
func unzip(src, dest string) error {
	r, err := zip.OpenReader(src)
	if err != nil {
		return err
	}
	defer func() {
		if err := r.Close(); err != nil {
			panic(err)
		}
	}()

	os.MkdirAll(dest, 0755)

	// Closure to address file descriptors issue with all the deferred .Close() methods
	extractAndWriteFile := func(f *zip.File) error {
		rc, err := f.Open()
		if err != nil {
			return err
		}
		defer func() {
			if err := rc.Close(); err != nil {
				panic(err)
			}
		}()

		path := filepath.Join(dest, f.Name)

		if f.FileInfo().IsDir() {
			os.MkdirAll(path, f.Mode())
		} else if f.FileInfo().Mode()&os.ModeSymlink != 0 {
			buffer := make([]byte, f.FileInfo().Size())
			size, err := rc.Read(buffer)
			if err != nil {
				return err
			}

			target := string(buffer[:size])

			err = os.Symlink(target, path)
			if err != nil {
				return err
			}
		} else {
			os.MkdirAll(filepath.Dir(path), f.Mode())
			f, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, f.Mode())
			if err != nil {
				return err
			}
			defer func() {
				if err = f.Close(); err != nil {
					panic(err)
				}
			}()

			_, err = io.Copy(f, rc)
			if err != nil {
				return err
			}
		}
		return nil
	}

	for _, f := range r.File {
		err := extractAndWriteFile(f)
		if err != nil {
			return err
		}
	}

	return nil
}
```

mholt:
Hmmm, want to make a pull request? :D

oaleynik:
@Skywalker13 really appreciate actual example - saves me a lot of time! If you will have an hour of free time or so, as @mholt said - PR would be awesome! Thanks again!

PR #40, opened 2017-08-25
Proper symlink support for Tar files

fkorotkov:
It appeared Tar implementation is not quite supporting archiving symlinks.

Added tests for archiving symlinks but I didn't manage to make them working for Zip archiver. Disabled testing Zip files(same as for Rar) to get a green build but open for suggestions on how to fix it for Zip file case. Not sure how symlinks are stored in a zip file. Seems there are no special instructions like for Tars.

Related to: https://github.com/mholt/archiver/issues/21


tar.go
```diff
@@ -145,6 +145,12 @@ func tarFile(tarWriter *tar.Writer, source, dest string) error {
 			header.Name += "/"
 		}
 
+		// special handling of symlinks
+		if header.Typeflag == tar.TypeSymlink {
+			linkDest, _ := os.Readlink(path)
```
mholt: This error should not be ignored; be sure to return it if there is one

archiver_test.go
```diff
@@ -17,6 +17,10 @@ func TestArchiver(t *testing.T) {
 			if _, ok := ar.(rarFormat); ok {
 				t.Skip("not supported")
 			}
+			// skip ZIP for now as well since it doesn't support symlinks
+			if _, ok := ar.(zipFormat); ok {
+				t.Skip("not supported")
```
mholt: Ummm, this is kind of a big deal, since zip is so popular and both archive/open functions are implemented (unlike RAR) - it would be good to find a better way to make the tests pass. :)

mholt:
@fkorotkov No longer interested in merging this?

fkorotkov:
@mholt unfortunately I spent too much time trying to make it work with Zip back then. As I recall I found out Zip doesn't really support symlinks. I ended up with a self-written version for my pet project. Maybe I'll come back to it once I'll need archiving for my bigger projects but at the moment I just don't have time to start investigating Zip use case again :-(

PR #46, opened 2017-10-11
fix tar for relative path directory

monotone:
when use `Tar` to make a relative path directory, it will make one more folder with the same name wrap for the the dest folder.

for example, the folder I want to make tar is:
```
- relative_path
  - a.txt
  - b.txt
```

after I use like this:

```go
archiver.Tar.Make(tarPath, []string{"./relative_path"})
```

when I try to untar it, I get:
```
- relative_path
 - relative_path
  - a.txt
  - b.txt
```

So, I add this code to make all path for tar is abs path to fix this.

### DIFF
--- a/archiver.go
+++ b/archiver.go
@@ -13,10 +13,14 @@ import (
 type Archiver interface {
 	// Match checks supported files
 	Match(filename string) bool
-	// Make makes an archive.
+	// Make makes an archive file on disk.
 	Make(destination string, sources []string) error
-	// Open extracts an archive.
+	// Open extracts an archive file on disk.
 	Open(source, destination string) error
+	// Write writes an archive to a Writer.
+	Write(output io.Writer, sources []string) error
+	// Read reads an archive from a Reader.
+	Read(input io.Reader, destination string) error
 }
 
 // SupportedFormats contains all supported archive formats
--- a/archiver.go
+++ b/archiver.go
@@ -35,6 +35,17 @@ func RegisterFormat(name string, format Archiver) {
 	SupportedFormats[name] = format
 }
 
+// MatchingFormat returns the first archive format that matches
+// the given file, or nil if there is no match
+func MatchingFormat(fpath string) Archiver {
+	for _, fmt := range SupportedFormats {
+		if fmt.Match(fpath) {
+			return fmt
+		}
+	}
+	return nil
+}
+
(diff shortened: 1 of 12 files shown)


mholt:
Thanks for this contribution... but I do have a question first: how are you untarring the file?

monotone:
Let me give a detail example:
This is the folder I want to tar:
```bash
ᐅ tree ./wrapper
./wrapper
├── pluginwrapper.go
└── pluginwrapper_test.go

0 directories, 2 files
```

And I run this code for it:
```go
func TestTar(t *testing.T) {
	tarPath := "./testdata/wrapper.tar"
	err := archiver.Tar.Make(tarPath, []string{"./wrapper"})
	if err != nil {
		t.Fatal(err)
	}

	err = archiver.Tar.Open(tarPath, "./testdata")
	if err != nil {
		t.Fatal(err)
	}
}
```

But I get this result:
```bash
ᐅ tree ./testdata
./testdata
├── wrapper
│   └── wrapper
│       ├── pluginwrapper.go
│       └── pluginwrapper_test.go
└── wrapper.tar
```

Where I expect is this : 
```bash
ᐅ tree ./testdata
./testdata
├── wrapper
│   ├── pluginwrapper.go
│   └── pluginwrapper_test.go
└── wrapper.tar
```

mholt:
Okay, I think I understand better now. Hmm, I'm not sure I understand _why_ that is happening in the first place.

Can you please add a test case to demonstrate that your change fixes it? (Make sure the test fails without your change.)

monotone:
I`m sorry, I have not thought about how to test.

I know it is work fine for taring with absolute path. So I think that is the easiest way to fix it, which make the path  always to be absolute path.

Just for this idea, I add some code to make all the path is absolute path when to tar it.

mholt:
Perhaps fixing how we strip the prefix is a better fix than converting everything to absolute paths?

Also, test cases can be built pretty easily by adding files to the testdata folder in such a way that the behavior is triggered.

monotone:
I know it was your original intention, though it was a failure. And I try to fix it with the same idea, and failed too.  

I will try it again.

monotone:
I have changed the method to fix it, and add some test case for it.

mholt:
I think this is fixed after #99.

PR #48, opened 2017-10-17
fix tar dest in src

monotone:
when i view the comment `our new tar file is inside the directory being archived; skip it` in function `tarFile` .

I found that when the `dest` is a relative path with prefix "./", it must not be equal with header.Name

Because the header.Name is always be a relative path without prefix "./".

So i test it, and find a method to fix it.  Just a amend all the path to be absolute path which will pass to function `tarFile`.  

I think that is really a good idea to make all the path to be absolute path , because any of them will be compared with the standard format path from the `WalkFunc`.

**I`am sorry for my poor english, it make me realy do`t not how to describe this problem clearly. May you could understand me**


tar_test.go
```diff
@@ -0,0 +1,201 @@
...
+			}
+
+			testSame(t, name, dest, outfile)
+		})
+	}
+}
+
+// symmetricTest compares the contents of a destination directory to the contents
+// of the test corpus and tests that they are equal.
+func testSame(t *testing.T, name, dest, ignore string) {
```
shurcooL: Function name `testSame` doesn't match the name `symmetricTest` used in documentation here.

monotone:
I amend the document already. thanks

weingart:
If I read this correctly, this puts absolute paths into the tar archive?

If so, this makes tar archives not portable to other locations within a filesystem, and likely not something we want to do.

weingart:
I do not believe that this will be a feasible approach.

PR #50, opened 2017-11-06
Add support to preserve UID and GID when extracting a tar* archive

T4cC0re:
Heyho,

I added the additional methods `OpenPreserve` and `ReadPreserve` that allow for keeping UID and GID in upon extraction.  
Additionally, they will set extended/special attributes such as SUID, SGID and sticky.
All that will have no effect on Windows platforms.  
`Rar` and `Zip` implementations have received a dummy method to keep the uniform interface. I decided to return an error. But maybe it would be desirable to transparently just ignore the 'preservation' and proxy the call to `Open`/`Read`?  
The private API in `Tar` changed, but the public API is kept compatible, so this should be easy to incorporate after a short review and approval from you :)

Cheers,

Hendrik

### DIFF
--- a/archiver.go
+++ b/archiver.go
@@ -9,8 +9,8 @@ import (
 	"runtime"
 )
 
-// Archiver represent a archive format
-type Archiver interface {
+// PreservableArchiver represent a archive format
+type PreservableArchiver interface {
 	// Match checks supported files
 	Match(filename string) bool
 	// Make makes an archive file on disk.
@@ -25,6 +25,20 @@ type Archiver interface {
 	OpenPreserve(source, destination string) error
 	// Read reads an archive from a Reader while preserving uid/gid.
 	ReadPreserve(input io.Reader, destination string) error
+}
+
+// Archiver represent a archive format
+type Archiver interface {
+	// Match checks supported files
+	Match(filename string) bool
+	// Make makes an archive file on disk.
+	Make(destination string, sources []string) error
+	// Open extracts an archive file on disk.
+	Open(source, destination string) error
+	// Write writes an archive to a Writer.
+	Write(output io.Writer, sources []string) error
+	// Read reads an archive from a Reader.
+	Read(input io.Reader, destination string) error
 }
 
 // SupportedFormats contains all supported archive formats
--- a/archiver.go
+++ b/archiver.go
@@ -21,6 +21,10 @@ type Archiver interface {
 	Write(output io.Writer, sources []string) error
 	// Read reads an archive from a Reader.
 	Read(input io.Reader, destination string) error
+	// Open extracts an archive file on disk while preserving uid/gid.
+	OpenPreserve(source, destination string) error
+	// Read reads an archive from a Reader while preserving uid/gid.
+	ReadPreserve(input io.Reader, destination string) error
 }
 
 // SupportedFormats contains all supported archive formats
--- a/tar.go
+++ b/tar.go
@@ -9,6 +9,8 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"runtime"
+	"syscall"
 )
 
 // Tar is for Tar format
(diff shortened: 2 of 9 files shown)


T4cC0re:
Btw: This has been tested on Linux and Windows (because of platform specific/exclusive code)

T4cC0re:
I just noticed, the dummy methods don’t make any sense whatsoever with the split interfaces. I’ll remove them later today.

T4cC0re:
I think I would recommend squash-merging this to keep a clean history. My commits are a bit back-and-forth

archiver.go
```diff
@@ -9,6 +9,24 @@ import (
 	"runtime"
 )
 
+// PreservableArchiver represent a archive format
+type PreservableArchiver interface {
+	// Match checks supported files
```
weingart: Embedding the Archiver interface into this one may be a better option.

archiver.go
```diff
@@ -9,6 +9,24 @@ import (
...
+	// Make makes an archive file on disk.
+	Make(destination string, sources []string) error
+	// Open extracts an archive file on disk.
+	Open(source, destination string) error
+	// Write writes an archive to a Writer.
+	Write(output io.Writer, sources []string) error
+	// Read reads an archive from a Reader.
+	Read(input io.Reader, destination string) error
+	// Open extracts an archive file on disk while preserving uid/gid.
+	OpenPreserve(source, destination string) error
```
weingart: I personally feel that extending this interface with more methods may be not required.  Having a few simple options that can be used to pass into Open() and Make() may do the trick.  Sort of something like:
```go
type Archiver interface {
  // ...
  Open(source, destination string, opts ...option)
  // ....
}

type option // ...
```

tar.go
```diff
@@ -210,25 +234,43 @@ func untar(tr *tar.Reader, destination string) error {
...
 
 // untarFile untars a single file from tr with header header into destination.
-func untarFile(tr *tar.Reader, header *tar.Header, destination string) error {
+func untarFile(tr *tar.Reader, header *tar.Header, destination string, preserveIds bool) error {
+	var err error
 	switch header.Typeflag {
 	case tar.TypeDir:
-		return mkdir(filepath.Join(destination, header.Name))
+		err = mkdir(filepath.Join(destination, header.Name))
+		break
```
weingart: break are not required

tar.go
```diff
@@ -210,25 +234,43 @@ func untar(tr *tar.Reader, destination string) error {
...
 	default:
-		return fmt.Errorf("%s: unknown type flag: %c", header.Name, header.Typeflag)
+		err = fmt.Errorf("%s: unknown type flag: %c", header.Name, header.Typeflag)
+		break
+	}
+	if err != nil {
+		return err
+	}
+
+	if preserveIds && header.Name != "./" && header.Name != "." && runtime.GOOS != "windows" {
```
weingart: What about if header.Name == "././."?  I'm not sure why you're trying to prevent the os.Lchown() for the current directory.

tar.go
```diff
@@ -210,25 +234,43 @@ func untar(tr *tar.Reader, destination string) error {
...
+		if err = os.Lchown(filepath.Join(destination, header.Name), header.Uid, header.Gid); err != nil {
+			return err
+		}
+		// If a special permission is set usage of the syscall is required.
+		return syscall.Chmod(filepath.Join(destination, header.Name), uint32(header.Mode))
```
weingart: One of the problems that invoking syscall.Chmod() and friends has, is that they indiscriminately will work on what you tell them to.  I can see two possible issues with the above:
  - if a previous entry in the archive was a softlink with a target to something outside the archive open tree, this later chmod(2) would end up changing the referenced file
  - the header.Name may be malicious, and point outside the destination tree, which if root uses this, could cause issues

I think for this PR to make headway, we would need to be much more careful with our paths, the cleaning of them, how we handle them, and have appropriate testing to make sure we're not going to end up with issues.

mholt:
Thanks for the review @weingart and for the contribution, however, with the pending changes, and the already existing merge conflict, I don't think there's much we can do with this PR now. So, closing it, and if this feature is still desired it would be worth opening an issue to discuss the best way to do it before putting in the work to actually do it, to save effort.

PR #55, opened 2018-01-27
Add Functionality to Replace Illegal File and Folder Name Characters on Windows

jwbrown77:
Hello,

I have recently been using this library to extract .tar.gz tarballs created on Linux systems to several client platforms.  While everything "just works" on Linux and macOS, I ran into an issue where Windows systems would not extract files or folders with illegal characters (on Windows).  In my case, it was namely the colon ( : ) character.

The error message is: ```The filename, directory name, or volume label syntax is incorrect.```

Screenshot of this behavior:

invalid_filename

After making the attached changes, which replace these illegal (on Windows) characters with underscore ( _ ), my Go program can now successfully extract these tarballs.

I'm not completely certain this is the "best" approach to this issue, as it assumes everyone wants the underscore as the replacement, and it incurs a performance penalty running the strings' "Replacer" method.  However, I figured I'd leave it here in case it helps the maintainers or anyone else.

Thanks for the great library.

- Jeremy

### DIFF
--- a/archiver.go
+++ b/archiver.go
@@ -5,8 +5,10 @@ import (
 	"io"
 	"log"
 	"os"
+	"path"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"strings"
 )
 
@@ -39,6 +41,9 @@ var windowsReplacer = strings.NewReplacer(
 	"|", "_",
 )
 
+// replacedDirs tracks the replaced directory name for any given directory
+var replacedDirs = make(map[string]string)
+
 // RegisterFormat adds a supported archive format
 func RegisterFormat(name string, format Archiver) {
 	if _, ok := SupportedFormats[name]; ok {
@@ -59,9 +64,60 @@ func MatchingFormat(fpath string) Archiver {
 	return nil
 }
 
+// replaceInvalidChars corrects invalid file and directory characters,
+// and in the case of collision with an existing file or directory,
+// adds an incrementing integer to the end
+func replaceInvalidChars(name string, dir bool, rtype *strings.Replacer) string {
+	rawName := name
+	name = rtype.Replace(name)
+
+	if rawName != name {
+		// if a file belongs in a directory whose name was already replaced due
+		// to invalid characters, replace the raw directory name with the
+		// previously assigned replacement name
+		if !dir {
+			newDir, ok := replacedDirs[filepath.Dir(rawName)]
+
+			if ok {
+				name = path.Join(newDir, filepath.Base(name))
+			}
+		}
+
+		// if the proposed replacement name doesn't yet exist, it is good to go
+		if _, err := os.Stat(name); os.IsNotExist(err) {
+			if dir {
+				replacedDirs[rawName] = name
+			}
+
+			return name
+		}
+
+		// otherwise, loop until a unique sequence is found
+		seq := 0
+
+		for {
+			seq++
+
+			nameSeq := name + "-" + strconv.Itoa(seq)
+
+			if _, err := os.Stat(nameSeq); err == nil {
+				continue
+			}
+
+			if dir {
+				replacedDirs[rawName] = nameSeq
+			}
+
+			return nameSeq
+		}
+	}
+
+	return name
+}
+
 func writeNewFile(fpath string, in io.Reader, fm os.FileMode) error {
 	if runtime.GOOS == "windows" {
-		fpath = windowsReplacer.Replace(fpath)
+		fpath = replaceInvalidChars(fpath, false, windowsReplacer)
 	}
 
 	err := os.MkdirAll(filepath.Dir(fpath), 0755)
@@ -89,8 +145,8 @@ func writeNewFile(fpath string, in io.Reader, fm os.FileMode) error {
 
 func writeNewSymbolicLink(fpath string, target string) error {
 	if runtime.GOOS == "windows" {
-		fpath = windowsReplacer.Replace(fpath)
-		target = windowsReplacer.Replace(target)
+		fpath = replaceInvalidChars(fpath, false, windowsReplacer)
+		target = replaceInvalidChars(target, false, windowsReplacer)
 	}
 
 	err := os.MkdirAll(filepath.Dir(fpath), 0755)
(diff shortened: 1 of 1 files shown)


archiver.go
```diff
@@ -26,6 +27,18 @@ type Archiver interface {
...
+var WindowsReplacer = strings.NewReplacer(
+	"<", "_",
+	">", "_",
+	":", "_",
+	"\"", "_",
+	"/", "_",
+	"?", "_",
+	"*", "_",
+	"|", "_",
+)
```
mholt: - I'm not sure this needs to be exported.

- We could just remove the character entirely, rather than replacing it.

- This will introduce filename collisions, for example `Hello?.txt` and `Hello:.txt` will become the same file. That might cause extractions to fail.

jwbrown77:
Thanks for the feedback Matt:

- You're absolutely right on the export, stupid mistake which merited immediate correction (change is committed 37414f5)
- Removing the illegal character instead of replacing is also valid.  I considered making the replacement character configurable, but that seemed over-engineered to me at the time.
- For the potential name collision - good point.  All that comes to mind is running a stat to see if the file already exists, and if so, use another character - but I suppose that would slow things down even further.  Any ideas?

mholt:
Thanks. I think the _proper_ way to handle the name collision would be to do the replacement, and _if_ a replacement was made, to see if that filename already exists. If so, to append a "-1" to it, and if that exists, a "-2", and so on. An optimization could be to store the last counter's value for a certain filename while extraction occurs, up to the last, say, 10 or 100 names or something. Or even just the last one (probably much simpler!), if we expect colliding names to happen together.

It's a little more involved but I think is most correct. What do you think?

jwbrown77:
Hi Matt,

Your idea sounds reasonable, and I think only running stat for a potential collision only after a replacement is performed is a sensible optimization.

Unfortunately, I don't think the "Replacer" method returns this information - but should be able to compare the raw input with the output.  Hopefully that isn't too slow.

I'll try to take a crack at this, though I may not get to it right away due to professional commitments.

Thanks for the constructive feedback.

Jeremy

jwbrown77:
Hi Matt,

My first attempt at adding the described sequence is committed in ebe6095 .  Every replacement now checks for a colliding file or directory name, and in case it is detected, will loop until it finds a unique sequence number to append, in the format of "-n" where n is the sequence.

In my testing, I created the following tarball on macOS:

```
tar tzf archiver-test.tar.gz
archiver-test/
archiver-test/dir</
archiver-test/test>.txt
archiver-test/test?.txt
archiver-test/dir>/
archiver-test/test<.txt
archiver-test/regular1.txt
archiver-test/regular2.txt
archiver-test/test:.txt
archiver-test/dir>/more:.txt
archiver-test/dir>/more<.txt
archiver-test/dir>/more>.txt
archiver-test/dir>/more?.txt
archiver-test/dir>/regular1.txt
archiver-test/dir>/regular2.txt
archiver-test/dir</less:.txt
archiver-test/dir</less<.txt
archiver-test/dir</less>.txt
archiver-test/dir</less?.txt
archiver-test/dir</regular1.txt
archiver-test/dir</regular2.txt
```

A test program to call this:

```go
package main

import (
	"fmt"

	"github.com/jwbrown77/archiver"
)

func main() {
	if err := archiver.TarGz.Open("archiver-test.tar.gz", "."); err != nil {
		fmt.Println(err)
	}
}
```

And received the following output:

```
cat windows.txt
Folder PATH listing
Volume serial number is 82C2-03F2
C:\USERS\USER\ARCHIVER-TEST
|   regular1.txt
|   regular2.txt
|   test_.txt
|   test_.txt-1
|   test_.txt-2
|   test_.txt-3
|
+---dir_
|       less_.txt
|       less_.txt-1
|       less_.txt-2
|       less_.txt-3
|       regular1.txt
|       regular2.txt
|
\---dir_-1
        more_.txt
        more_.txt-1
        more_.txt-2
        more_.txt-3
        regular1.txt
        regular2.txt
```

The idea being: given the original directory names of ```dir<``` and ```dir>``` - name test files with invalid characters to start with "less..." and "more..." to indicate where the final output should have been stored.

I tried to keep the function's naming scheme generic to potentially be able to reuse it for replacers other than Windows in the future - but this may such an edge case that such naming is unnecessary.

Anyway, let me know if you have any feedback on this.

Outside of that, I think the last item is the question to whether or not the invalid character should be replaced with underscore, nothing at all (removed), or something else.  I personally like the underscore to delineate, but acknowledge it's arbitrary - and it's your library, so I'm happy to change it to anything of your choosing.

Thanks,

Jeremy

mholt:
Thanks Jeremy -- I haven't forgotten about this, but things got a little crazy this week so I probably won't get to this immediately. (Using the underscore is fine, though!)

I do look forward to going over this, it seems to be a very high quality submission. There's some tests failing on Windows right now, maybe that could be investigated in the meantime.

jwbrown77:
Hi Matt,

No worries on timeline.

I'll write a unit test for the replaceInvalidChars function and submit soon (should have done this from the start).

Thanks,

Jeremy

jwbrown77:
@mholt - I just wanted to let you know I haven't forgotten about this.  Work has been hectic lately and I haven't had as much time as I'd like to spend on this.

Testing has turned up some issues with this implementation, namely "C:\", for example, gets turned into "C_\" - which has a completely different meaning.

Need some time to research how to best approach this...

Thanks for your feedback.

Jeremy

mholt:
No problem, I'm booked up for at least a week as well. Thanks!

mholt:
Not sure we've reached a totally satisfactory solution to the problem, other than to not write archives that have crazy filenames (and preferably to just not use Windows, tbh) -- considering certain filenames aren't allowed either, like CON or AUX. ([It's that famous 40+ year old bug.](https://twitter.com/foone/status/1058676834940776450))

Anyway, thanks for your time, but given the merge conflict and the pending rewrite of the core I think I'm going to have to close this. With the rewrite, errors are optional non-fatal, so malformed filenames won't be a showstopper.

jwbrown77:
Thanks @mholt and sorry.  I did try to correct your (valid) concerns, but got stuck with how Windows handles drive letters vs. *nix style mount points.

My initial patch is imperfect but it fits my need so I'm fine continuing to use locally.

If anyone else hits this and you feel it's important, reach out and I'll take another swing at it.

PR #63, opened 2018-03-30
add filtering folders and files

bingyangzeng:
add ./conf/app.ini
example:
[filters]
path = .git __MACOSX


mholt:
Thanks for the contribution, but I think we should discuss new features in issues first -- also the tests will need to pass. Feel free to open an issue to discuss this!

Issue #62, opened 2018-03-28
Compressing tarball on Windows


addelovein:
When compressing a entire folder with 
archiver.TarGz.Make("myfolder.tar.gz", []string{"./myfolder"})

If i open that arhive in Windows it will look great i can go through the directory structure... However when i extract this on linux it wil create filenames like this:
"picons\picons\france\1_0_1_EB2_AAFF_7EB6_0_0_0_0.png"
and empty folders named this:
"picons\picons\france"

That is not paths that is the actual name of the file....

halfcrazy:
https://github.com/mholt/archiver/blob/master/tar.go#L149

Maybe need another pr like https://github.com/mholt/archiver/pull/37 to handle this for tar format.

halfcrazy:
@addelovein I have a quick implementation at #64 , could you test on windows?

addelovein:
@halfcrazy Sorry for delay but thanks for the attempt... However still result when extracting on NIX is:
drwxrwxrwx    2 root     root            40 May  1 17:00 picons\picons\united_states
-rw-rw-rw-    1 root     root          5724 May  1 16:59 picons\picons\united_states\1_0_1_2331_AAFF_7EB6_0_0_0_0.png
-rw-rw-rw-    1 root     root          3601 May  1 16:59 picons\picons\united_states\1_0_1_2332_AAFF_7EB6_0_0_0_0.png
-rw-rw-rw-    1 root     root          3174 May  1 16:59 picons\picons\united_states\1_0_1_2334_AAFF_7EB6_0_0_0_0.png
-rw-rw-rw-    1 root     root          6451 May  1 16:59 picons\picons\united_states\1_0_1_2338_AAFF_7EB6_0_0_0_0.png

johnarok:
@addelovein can you re-test? With pr #77 merged, this should be fixed now.

johnarok:
Closing since no activity.

PR #64, opened 2018-04-17
[WIP]Fix compressing tarball on Windows

halfcrazy:
Regards to tar spec, name in header should contains only slash

Fix https://github.com/mholt/archiver/issues/62

From http://www.gnu.org/software/tar/manual/html_node/Standard.html
> The name field is the file name of the file, with directory names (if any) preceding the file name, separated by slashes.

Need test.

### DIFF
--- a/tar.go
+++ b/tar.go
@@ -146,7 +146,7 @@ func tarFile(tarWriter *tar.Writer, source, dest string) error {
 		}
 
 		if baseDir != "" {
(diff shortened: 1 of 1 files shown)


mholt:
@halfcrazy Any interest in finishing this PR? If not, we can close it.

halfcrazy:
@mholt close #64 Will create another pr later.

Issue #27, opened 2017-02-25
Error when attempting to extract from git generated gzipped tarball


JackMordaunt:
Hi there.

I have a gzipped tarball generated from git (specifically using the github API).
When I attempt to Open it using `archiver.TarGz.Open` I get this error: `"pax_global_header: unkown type flag: g"`

I searched for "pax_global_header" and I learned it seems to be an extended header that holds the git commit ID. I'm running OSX 10.12.

How should I go about investigating further? Is this known? Has anyone else had an issue?

Thanks for the nice utility.
Cheers!

mholt:
Hey Jack, I'm not sure about it. The error probably comes from this line: https://github.com/mholt/archiver/blob/c55f48c22890807c9698370fe5e08eca3704834b/tar.go#L213 - in that we don't handle that type flag.

How is that type of header handled? Can we just ignore it for our purposes?

svnpenn:
@mholt this is a problem, because this program cannot handle archives that are
hosted at this repository. You say on the initial commit in 2016:

> Package archiver makes it trivially easy to make .zip and .tar.gz files

http://github.com/mholt/archiver/blame/e08da12/README.md#L4

but if you download this file:

http://github.com/mholt/archiver/archive/v2.0.tar.gz

but this file cannot be opened by archiver:

~~~
$ ./archiver open archiver-2.0.tar.gz
pax_global_header: unknown type flag: g
~~~

where 7-Zip can handle it just fine:

~~~
$ ./7za x archiver-2.0.tar.gz
7-Zip (A) 9.20  Copyright (c) 1999-2010 Igor Pavlov  2010-11-18
Processing archive: archiver-2.0.tar.gz
Extracting  archiver-2.0.tar
Everything is Ok
Size:       61440
Compressed: 17703
~~~

JackMordaunt:
@mholt Yes I believe the header can be ignored, at least for the vast majority of users downloading archives from GitHub.

I'm not sure how the header is handled.

InTheCloudDan:
is there any workaround for this issue? I'm looking to use archiver library with a tarball from github and running into this exact problem.

PR #66, opened 2018-04-30
This commit effectively ignores PAX Header

InTheCloudDan:
From #27 this will ignore the PAX Header. I ran `go test` but not sure if there are additional tests to run? Locally this did fix my issue of being able to untar a tarball from github.


mholt:
@InTheCloudDan Thanks for the contribution! Do you think you could add a comment and, if practical, a test to verify this is the correct behavior? Then I will look at merging it in.

PR #79, opened 2018-08-25
Fixes #57 by exiting with 0 when -h flag is used

### LINKED ISSUE #57: Binary invoked with -h exits with non-zero exit status
./archiver_linux_amd64 -h returns a non-zero exit status. It should be zero.

johnarok:


johnarok:
For now, I have added a line to the usage on the -h flag, If we see a need for more formal help in future we can consider adding a command/using flag package

PR #83, opened 2018-09-02
Updated tar archiver to create directories similar to zip archiver

payaaam:
changes: _patch_

**Bug Fix**
I noticed that when I create tar files, the output directory structure has nesting that is not intended. I then tested the zip archiver and saw that this was not an issue. I updated the tar archiver to have same directory parsing logic as zip.

**Changes**
* updated path to fpath in filepath.Walk to allow for path package import
* Updated baseDir handling to match zip

### DIFF
--- a/tar.go
+++ b/tar.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path"
 	"path/filepath"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


tar.go
```diff
@@ -135,18 +136,22 @@ func tarFile(tarWriter *tar.Writer, source, dest string) error {
...
-		header, err := tar.FileInfoHeader(info, path)
+		header, err := tar.FileInfoHeader(info, fpath)
 		if err != nil {
-			return fmt.Errorf("%s: making header: %v", path, err)
+			return fmt.Errorf("%s: making header: %v", fpath, err)
 		}
 
 		if baseDir != "" {
-			header.Name = filepath.ToSlash(filepath.Join(baseDir, strings.TrimPrefix(path, source)))
+			name, err := filepath.Rel(source, fpath)
```
weingart: Any chance you can add a test for this here, that fails without this diff, and passes with this diff?

mholt:
This might not be an issue anymore with #99. Will probably close this if there's no more activity or unless someone confirms it's still an issue.

PR #89, opened 2018-09-11
Fixes #22 by checking if file exists at destination before writing

### LINKED ISSUE #22: Support for avoiding overwrite upon opening archive
Hi @mholt, this is a great tool/library! 

I've found it useful to avoid overwriting files if they already exist when opening an archive to a folder. I'm wondering if you'd like to incorporate this as an option to your library? Here's my change: https://github.com/schollz/archiver/commit/8b912cafcb94be510e20d5c27ac803e3dd978de3. Of course, it would need to be amended so it can be toggled as optional.

johnarok:
Thanks to @schollz for initial work

### DIFF
--- a/archiver.go
+++ b/archiver.go
@@ -48,6 +48,10 @@ func MatchingFormat(fpath string) Archiver {
 }
 
 func writeNewFile(fpath string, in io.Reader, fm os.FileMode) error {
+	if fileExists(fpath) {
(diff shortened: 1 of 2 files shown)


mholt:
File overwrite behavior is being improved (and made configurable) in #99, so I think this PR will be superseded.

PR #91, opened 2018-09-29
able to set sanity interface, cause someone may still want to extract…

wyxloading:
add DefaultExtractPathFunc can be set to self defined implement.
Main purpose is to support user defined behavior when meeting zip slip, or meeting situation that filename in zip not in utf8 encoding.

### DIFF
--- a/rar.go
+++ b/rar.go
@@ -88,7 +88,7 @@ func extract(rr *rardecode.Reader, destination string) error {
 			return err
 		}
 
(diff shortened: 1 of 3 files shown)


mholt:
Thanks for the contribution, but this is going to be superseded by #99, which implements functions for walking an archive, so you can decide before extracting what you want to do with it. Would you please take a look and see if any of the Walk functions will meet your needs?

wyxloading:
Great! I will take a look at #99

PR #60, opened 2018-03-07
preserve symlinks (fixes #21)

peterpostmann:
fix https://github.com/mholt/archiver/issues/21

PR #92, opened 2018-10-01
Add support for storing symlinks in tar and zip archives

### LINKED ISSUE #21: The symbolic links are not preserved when unpacking a zip (other formats not tested)

### LINKED ISSUE #31: symlinks  target file err fo tar
tar.go 131
header, err := tar.FileInfoHeader(info, path) 
----

link, e:= os.Readlink(path)
if e ==nil { path=link }
header, err := tar.FileInfoHeader(info, path)

### LINKED ISSUE #74: Allow relative symlinks
At the moment my relative symlinks are converted to absolute symlinks when using TarGz, this breaks the code I'm archiving.

jandubois:
Also implement extraction of symlinks from zip archives.

This PR also adds a relative symlink in the testdata directory. It passes all tests on OS X, but has not been tested on Windows.

This PR is a superset of changes from the following issues and PRs:

Fixes #21 
Fixes #31 
Fixes #60 
Fixes #74

### DIFF
--- a/tar.go
+++ b/tar.go
@@ -340,16 +340,16 @@ func (t *Tar) Write(f File) error {
 		return fmt.Errorf("missing file name")
 	}
 
-	filename := f.Name()
+	var linkTarget string
 	if isSymlink(f) {
 		var err error
-		filename, err = os.Readlink(f.Name())
+		linkTarget, err = os.Readlink(f.Name())
 		if err != nil {
 			return fmt.Errorf("%s: readlink: %v", f.Name(), err)
 		}
 	}
 
-	hdr, err := tar.FileInfoHeader(f, filepath.ToSlash(filename))
+	hdr, err := tar.FileInfoHeader(f, filepath.ToSlash(linkTarget))
 	if err != nil {
 		return fmt.Errorf("%s: making header: %v", f.Name(), err)
 	}
--- a/tar.go
+++ b/tar.go
@@ -285,12 +285,14 @@ func (t *Tar) writeWalk(source, topLevelFolder, destination string) error {
 			return handleErr(err)
 		}
 
-		file, err := os.Open(fpath)
-		if err != nil {
-			return handleErr(fmt.Errorf("%s: opening: %v", fpath, err))
+		var file io.ReadCloser
+		if info.Mode().IsRegular() {
+			file, err = os.Open(fpath)
+			if err != nil {
+				return handleErr(fmt.Errorf("%s: opening: %v", fpath, err))
+			}
+			defer file.Close()
 		}
-		defer file.Close()
-
 		err = t.Write(File{
 			FileInfo: FileInfo{
 				FileInfo:   info,
@@ -338,7 +340,15 @@ func (t *Tar) Write(f File) error {
 		return fmt.Errorf("missing file name")
 	}
 
-	hdr, err := tar.FileInfoHeader(f, f.Name())
+	var linkTarget string
+	if (f.Mode() & os.ModeSymlink) != 0 {
+		var err error
+		linkTarget, err = os.Readlink(f.Name())
+		if err != nil {
+			return fmt.Errorf("%s: readlink: %v", f.Name(), err)
+		}
+	}
+	hdr, err := tar.FileInfoHeader(f, filepath.ToSlash(linkTarget))
 	if err != nil {
 		return fmt.Errorf("%s: making header: %v", f.Name(), err)
 	}
--- a/tar.go
+++ b/tar.go
@@ -340,15 +340,16 @@ func (t *Tar) Write(f File) error {
 		return fmt.Errorf("missing file name")
 	}
 
-	var linkTarget string
-	if (f.Mode() & os.ModeSymlink) != 0 {
+	filename := f.Name()
+	if isSymlink(f) {
 		var err error
-		linkTarget, err = os.Readlink(f.Name())
+		filename, err = os.Readlink(f.Name())
 		if err != nil {
 			return fmt.Errorf("%s: readlink: %v", f.Name(), err)
 		}
 	}
-	hdr, err := tar.FileInfoHeader(f, filepath.ToSlash(linkTarget))
+
+	hdr, err := tar.FileInfoHeader(f, filepath.ToSlash(filename))
 	if err != nil {
 		return fmt.Errorf("%s: making header: %v", f.Name(), err)
 	}
@@ -359,7 +360,7 @@ func (t *Tar) Write(f File) error {
 	}
 
 	if f.IsDir() {
-		return nil
+		return nil // directories have no contents
 	}
 
 	if hdr.Typeflag == tar.TypeReg {
--- a/zip.go
+++ b/zip.go
@@ -329,11 +329,8 @@ func (z *Zip) Write(f File) error {
 		return fmt.Errorf("%s: making header: %v", f.Name(), err)
 	}
 
-	if f.IsDir() {
-		return nil
-	}
-
-	if (header.Mode() & os.ModeSymlink) != 0 {
+	switch header.Mode() & os.ModeType {
+	case os.ModeSymlink:
 		linkTarget, err := os.Readlink(f.Name())
 		if err != nil {
 			return fmt.Errorf("%s: readlink: %v", f.Name(), err)
@@ -342,10 +339,7 @@ func (z *Zip) Write(f File) error {
 		if err != nil {
 			return fmt.Errorf("%s: writing symlink target: %v", f.Name(), err)
 		}
-		return nil
-	}
-
-	if header.Mode().IsRegular() {
+	case 0: // regular file
 		if f.ReadCloser == nil {
 			return fmt.Errorf("%s: no way to read file contents", f.Name())
 		}
--- a/zip.go
+++ b/zip.go
@@ -186,15 +186,6 @@ func (z *Zip) extractNext(to string) error {
 	if !ok {
 		return fmt.Errorf("expected header to be zip.FileHeader but was %T", f.Header)
 	}
-	if (header.FileInfo().Mode() & os.ModeSymlink) != 0 {
-		buffer := make([]byte, header.FileInfo().Size())
-		size, err := f.Read(buffer)
-		if err != nil && err != io.EOF {
-			return fmt.Errorf("%s: reading symlink target: %v", header.Name, err)
-		}
-		linkTarget := string(buffer[:size])
-		return writeNewSymbolicLink(filepath.Join(to, header.Name), linkTarget)
-	}
 	return z.extractFile(f, filepath.Join(to, header.Name))
 }
 
@@ -207,6 +198,20 @@ func (z *Zip) extractFile(f File, to string) error {
 	// do not overwrite existing files, if configured
 	if !z.OverwriteExisting && fileExists(to) {
 		return fmt.Errorf("file already exists: %s", to)
+	}
+
+	header, ok := f.Header.(zip.FileHeader)
+	if !ok {
+		return fmt.Errorf("expected header to be zip.FileHeader but was %T", f.Header)
+	}
+	if (header.FileInfo().Mode() & os.ModeSymlink) != 0 {
+		buffer := make([]byte, header.FileInfo().Size())
+		size, err := f.Read(buffer)
+		if err != nil && err != io.EOF {
+			return fmt.Errorf("%s: reading symlink target: %v", header.Name, err)
+		}
+		linkTarget := string(buffer[:size])
+		return writeNewSymbolicLink(to, linkTarget)
 	}
 
 	return writeNewFile(to, f, f.Mode())
--- a/zip.go
+++ b/zip.go
@@ -186,6 +186,15 @@ func (z *Zip) extractNext(to string) error {
 	if !ok {
 		return fmt.Errorf("expected header to be zip.FileHeader but was %T", f.Header)
 	}
+	if (header.FileInfo().Mode() & os.ModeSymlink) != 0 {
+		buffer := make([]byte, header.FileInfo().Size())
+		size, err := f.Read(buffer)
+		if err != nil && err != io.EOF {
+			return fmt.Errorf("%s: reading symlink target: %v", header.Name, err)
+		}
+		linkTarget := string(buffer[:size])
+		return writeNewSymbolicLink(filepath.Join(to, header.Name), linkTarget)
+	}
 	return z.extractFile(f, filepath.Join(to, header.Name))
 }
 
@@ -244,12 +253,14 @@ func (z *Zip) writeWalk(source, topLevelFolder, destination string) error {
 			return handleErr(err)
 		}
 
-		file, err := os.Open(fpath)
-		if err != nil {
-			return handleErr(fmt.Errorf("%s: opening: %v", fpath, err))
+		var file io.ReadCloser
+		if info.Mode().IsRegular() {
+			file, err = os.Open(fpath)
+			if err != nil {
+				return handleErr(fmt.Errorf("%s: opening: %v", fpath, err))
+			}
+			defer file.Close()
 		}
-		defer file.Close()
-
 		err = z.Write(File{
 			FileInfo: FileInfo{
 				FileInfo:   info,
@@ -314,6 +325,18 @@ func (z *Zip) Write(f File) error {
 	}
 
 	if f.IsDir() {
+		return nil
+	}
+
+	if (header.Mode() & os.ModeSymlink) != 0 {
+		linkTarget, err := os.Readlink(f.Name())
+		if err != nil {
+			return fmt.Errorf("%s: readlink: %v", f.Name(), err)
+		}
+		_, err = writer.Write([]byte(filepath.ToSlash(linkTarget)))
+		if err != nil {
+			return fmt.Errorf("%s: writing symlink target: %v", f.Name(), err)
+		}
 		return nil
 	}
 
--- a/zip.go
+++ b/zip.go
@@ -182,14 +182,17 @@ func (z *Zip) extractNext(to string) error {
 		return err // don't wrap error; calling loop must break on io.EOF
 	}
 	defer f.Close()
+	return z.extractFile(f, to)
+}
+
+func (z *Zip) extractFile(f File, to string) error {
 	header, ok := f.Header.(zip.FileHeader)
 	if !ok {
 		return fmt.Errorf("expected header to be zip.FileHeader but was %T", f.Header)
 	}
-	return z.extractFile(f, filepath.Join(to, header.Name))
-}
 
-func (z *Zip) extractFile(f File, to string) error {
+	to = filepath.Join(to, header.Name)
+
 	// if a directory, no content; simply make the directory and return
 	if f.IsDir() {
 		return mkdir(to)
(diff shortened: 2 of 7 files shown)


mholt:
Nice! Looking forward to going through this, unless another collaborator beats me to it.

johnarok:
Adding my notes from testing using the archiver cli

Windows to Windows, Zip format -> Symlinks are **not** preserved
Windows to Linux, Zip format -> Symlinks are **not** preserved
Windows to Windows, Tar format -> Symlinks are preserved (**breaks** symlink with absolute path)
Windows to Linux, Tar format -> Symlinks are preserved (**breaks** symlink with absolute path)
```
lrwxrwxrwx 1 user01 user01 29 Oct 23 20:42 hello-link -> 'testdata\hello-dir\hello-link'
```

Linux to Linux, Zip format -> Symlinks are preserved (retains relative path)
Linux to Windows, Zip format -> Symlinks are **not** preserved
Linux to Linux, Tar format > Symlinks are preserved (retains relative path)
Linux to Windows, Tar format -> Symlinks are preserved (retains relative path)

jandubois:
Thanks for your comprehensive tests, @johnarok!

I'm travelling right now and only have access to a Macbook for the next 2 weeks, so can't do any Windows testing myself. I assume some of the failures are due to not normalizing the symlink paths to always use forward slashes for the path separator. I added a commit that calls `filepath.ToSlash` on all symlinks before storing them into an archive.

If you have the time, could you redo your tests with the updated PR?

I'm somewhat surprised that the appveyor tests didn't see the windows-to-windows failures you were seeing; did the tests pass for you on Windows? That would mean that the tests are not comprehensive enough. The only explanation I could see is if git didn't re-create the symlinks in the testdata directory.

johnarok:
Hi! @jandubois , so the windows tests indeed pass, the reason being a git clone on windows (`git version 2.16.2.windows.1` and `Windows 10` ) creates the symlink files as normal files. I am hoping at the end of the PR, we can add a test case to workaround this.

I retested by manually creating the symbolic link for the `proverb3.txt` file.

```
mklink proverb3.txt .\proverbs\extra\proverb3.txt
```

Below is what I observed this time

Windows to Windows, Zip format -> Symlinks are **not** preserved (they appear as normal files)
Windows to Linux, Zip format -> Symlinks are **not** preserved, Archiver CLI **fails**.
```
./archiver open testdata-windows.zip 
/home/user01/Projects/go/gowork/src/github.com/mholt/archiver/test/testdata/proverb3.txt: making symbolic link for: symlink  /home/user01/Projects/go/gowork/src/github.com/mholt/archiver/test/testdata/proverb3.txt: no such file or directory
```
Windows to Windows, Tar format -> Symlinks are preserved (**breaks** symlink with incorrect path)
```
10/24/2018  10:46 PM    <SYMLINK>      proverb3.txt [testdata\proverb3.txt]
```
Windows to Linux, Tar format -> Symlinks are preserved (**breaks** symlink with incorrect path)
```
lrwxrwxrwx 1 user01 user01   21 Oct 24 22:53 proverb3.txt -> 'testdata\proverb3.txt'
```
Linux to Linux, Zip format -> Symlinks are preserved (retains relative path)
Linux to Windows, Zip format -> Symlinks are **not** preserved (they appear as normal files)
Linux to Linux, Tar format > Symlinks are preserved (retains relative path)
Linux to Windows, Tar format -> Symlinks are preserved (retains relative path)

jandubois:
Hi @johnarok! I think I will need access to my Windows VMs to investigate this further, so there won't be any further activity from me for the next 7-10 days. I'll get back to you once I actually have run this on Windows myself.

Just for clarification, what are the expectations for older Windows versions that don't have proper symlink support? Or for versions that require admin permissions to create symlinks when the tests are being run as a normal user?

johnarok:
Hi @jandubois, I did some cursory research on the state of symlinks for windows in other projects, links [here](https://gist.github.com/johnarok/2babf8baa9dffa3201ed009a1d3e7ae4). 

_What are the expectations for older Windows versions that don't have proper symlink support?_ 
IMO, We should fail safe, that is error out if we cannot create the symlink and potentially allow the user to explicitly force  via a FLAG or ENV, to fallback - restore the files as normal files. /cc @mholt for his opinion on this. This would be different from what git is doing, which is fall back by default.

_Or for versions that require admin permissions to create symlinks when the tests are being run as a normal user?_
IMO, The tests are run as administrator in `appveyor`, so symlinks should work, but we need add a step to appveyor to set git config `core.symlinks=true`  to allow the test data to be cloned correctly i.e create the links. I was able to get this behavior on Windows 10.

Also, retested it with 1.11.2 go version and the results did not change much. I have uploaded the artifacts [here](https://github.com/johnarok/archiver-testing/tree/pr92) for reference.

mholt:
:-/ Well this complicated, isn't it. Thanks for the work, both of you!

I'd like to see this get to a state where it can be merged, but my PR #99 will have to come first, and it does not address the symlink issues, which means that this PR will have to be rebased -- fortunately, I think this PR should be easier to rebase.

FWIW the new code in #99 will allow the user to decide whether they want to continue on errors, so it's OK if there are errors, it doesn't have to stop the operation.

As for older versions of Windows -- I say drop support. I'm not thrilled with supporting old OSes. Especially when it makes things simpler.

So, I might be a little hands-off on this PR -- but don't let me slow you down. Go ahead and both of you do what you think is best and let's merge it in after my rewrite is finalized. In the meantime, you might want to rebase this fork off my `rewrite` branch...

mholt:
Okay, the rewrite is done! If you want to update this PR, go for it.

jandubois:
@mholt I've updated this PR on top of latest `master`, but haven't gotten around to looking into the Windows issues yet.

jandubois:
@johnarok I cannot reproduce your test failures. I'm also somewhat confused that your re-testing with the updated PR still showed a symlink with backslashes in them; that should have been fixed by that time.

Anyways, with the current state of the PR I used a Windows 10 x64 VM with the latest Windows updates, and I've enabled "Developer Mode" to be able to create symlinks without getting UAC prompts.

I installed these 2 packages to get `git` and `go`:

https://github.com/git-for-windows/git/releases/download/v2.19.1.windows.1/Git-2.19.1-64-bit.exe
https://dl.google.com/go/go1.11.2.windows-amd64.msi

I enabled `git config core.symlinks true` and switched to my PR branch. The directory has the expected symlinks:

```
 Directory of C:\Users\Jan\go\src\github.com\jandubois\archiver\testdata

11/20/2018  12:47 PM    <DIR>          .
11/20/2018  12:47 PM    <DIR>          ..
11/20/2018  12:41 PM             8,944 already-compressed.jpg
11/20/2018  12:47 PM    <SYMLINK>      exist [C:\target\does\not\exist]
11/20/2018  12:46 PM    <SYMLINK>      proverb3.txt [proverbs\extra\proverb3.txt]
11/20/2018  12:41 PM    <DIR>          proverbs
11/20/2018  12:41 PM                59 quote1.txt
```

All tests pass (I've temporarily patched `cmd/arc/main.go` to use `jandubois/archiver` for testing):

```
C:\Users\Jan\go\src\github.com\jandubois\archiver>go test
PASS
ok      github.com/jandubois/archiver   0.840s
```

I created 2 test archives:

```
arc archive foo.tar.gz testdata
arc archive foo.zip testdata
```

Both archives could be unpacked with `arc` again, both on Windows and on OS X, with the symlinks intact.

I could also unpack them on OS X with `tar xfvz` and `unzip`, again with the symlinks intact.

What did *not* work was unpacking them on Windows using the corresponding MinGW tools. I assume they don't support symlinks, or at least not by default. I believe their limitation is irrelevant to this PR.

I've also created the same tarball and zip files using `arc` on OS X and successfully unpacked them on Windows with the symlinks intact.

To me this means that the code is supporting symlinks on Windows as intended.

@johnarok Please provide more detailed instructions to reproduce your failures, if you believe they still need to be addressed!

The only open question in my mind is if the drive specified should be stripped from a full path symlink, i.e. store `/target/does/not/exist` instead of `C:/target/does/not/exist`. I feel that stripping the volume id is the right thing to do, but can't really provide a strong rationale for it.

Unrelated observation:

When using `unzip foo.zip` on OS X, using the zip file created on Windows, the `proverbs/` subdirectory has the wrong permissions (world writable):

```
drwxrwxrwx  5 jan  staff   160 20 Nov 12:41 proverbs
```

This does not happen with the zip file created on OS X, and doesn't happen with either zip file when using `arc unarchive`:

```
drwxr-xr-x  5 jan  staff   160 30 Sep 11:24 proverbs
```

I haven't looked further into it, as it is completely unrelated to the symlink code. It should probably get its own Github issue, but maybe you can double-check this when verifying this latest PR version?

jandubois:
Out of curiosity I just disabled "Developer Mode" and ran the tests again:

```
C:\Users\Jan\go\src\github.com\jandubois\archiver>go test
--- FAIL: TestArchiveUnarchive (0.02s)
    archiver_test.go:280: [zip] extracting archive [C:\Users\Jan\AppData\Local\Temp\archiver_test364499367\archiver_test.zip -> C:\Users\Jan\AppData\Local\Temp\archiver_test364499367\extraction_test_zip]: didn't expect an error, but got: reading file in zip archive: C:\Users\Jan\AppData\Local\Temp\archiver_test364499367\extraction_test_zip\testdata\exist: making symbolic link for: symlink C:\target\does\not\exist C:\Users\Jan\AppData\Local\Temp\archiver_test364499367\extraction_test_zip\testdata\exist: A required privilege is not held by the client.
FAIL
exit status 1
FAIL    github.com/jandubois/archiver   0.572s
```

That seems fine to me as well...

johnarok:
@jandubois, thank you for validating, I did not use the developer mode. I will pull the latest and test again to confirm - hopefully next week after the holidays.

mholt:
Something to consider: should how symlinks are handled be configurable? It's possible now. Maybe some users want to preserve them, while others want to follow/dereference them...

jandubois:
@johnarok I've added the appveyor config you requested. Tests are still passing, but I can't tell if appveyor actually does support symlinks or not (they may be running a version of git that ignores that option).

My thoughts on making symlink support configurable:

* For extraction I don't see any need for configuration: if the archive contains a symlink, it should be extracted as a symlink. If this isn't possible, then extraction should fail. The option to skip errors should be enough to deal with this.

* For creating archives, having an option to follow symlinks vs. storing them verbatim makes sense. The question then becomes: "What is the default behavior?"

I agree with @johnarok that the current behavior implemented by this PR is the desirable default. Only concern is that previously creating ZIP archives would have followed symlinks (I think, I haven't tested), and maintaining backwards compatibility is a good thing.

For TAR archives the code before this PR used to store the full path to the original file as a symlink, which is totally broken and not worth preserving.

My preferred way forward would be to switch the version to 4.0 and change the default mode for ZIP archive creation.

Adding an option for following symlinks is independent functionality and should be requested and/or implemented in a separate issue/pr.

mholt:
@jandubois I think I agree with almost all of what you've proposed!

The `archiver.go` file has two functions which I kept from the old versions:

- `writeNewHardLink` https://sourcegraph.com/github.com/mholt/archiver@d572b2e8b82726cee9476d1b9d63a7fe9b601ff1/-/blob/archiver.go#L318:1

- `writeNewSymbolicLink` https://sourcegraph.com/github.com/mholt/archiver@d572b2e8b82726cee9476d1b9d63a7fe9b601ff1/-/blob/archiver.go#L304

As you can see there, the only references to those functions are in writing `tar` archives. I'm not sure if they're handled correctly, I haven't tested them with the v3 code that we currently have.

In short, I'd love to see this package brought up to speed with a standard/consistent way of handling symlinks (and hard links maybe? doesn't have to be in this PR).

So far, I think we've decided:

- Handling of symlinks should be configurable. This is now possible as a struct field on the relevant types: `Zip` and `Tar`, mostly. It would only apply to creation?

- Symlinks contained in archives should be extracted as symlinks, at least as the default behavior. Makes sense to me.

Not too concerned about backwards compatibility. I think we pretty much broke symlink support with v3, so I don't think that skipping or adding it would be breaking either way. We might be able to keep it in the v3 tree.

So, with that said, how should we proceed on this PR? What else needs to be done?

jandubois:
This PR uses `writeNewSymbolicLink` for both TAR and ZIP files consistently, so I think that is already "up to speed".

I'm not sure if preserving hard links is an important goal. It is not supported by ZIP at all, afaik.

Even with tarballs it is a problematic feature, e.g. you cannot selectively extract a later hardlink unless you have extracted the file under the first hardlink name already.  Hardlink support may also be platform specific; I don't know if there is module that provides a device/inode view of Windows filesystems.

So to me it feels somewhat at odds with the project value of

> It is not meant to be a replacement for specific archive format tools like tar, zip, etc. that have lots of features and customizability.

Either way, supporting hardlinks would be completely different from softlinks, so should be a separate issue. Personally I feel it is not worth the effort.

> So, with that said, how should we proceed on this PR? What else needs to be done?

If you are fine with making the current behavior the default, then I think this PR should be ready for merging, and a new Github issue should be opened to add an option to dereference symbolic links while creating archives.

I'm personally not interested in that feature, so probably wouldn't want to implement it: the change itself is rather straight-forward, but I'm not sure how you would want to re-arrange the tests to accommodate testing different option settings. Once the test framework for this is in place, I would be happy to add the conditionals to the actual code itself though. 😄

mholt:
Excellent. You're right, hard links are probably not a priority at all, and maybe should even be removed entirely from the code. As for this PR currently, I will begin a review now.

Just to be sure I understand clearly going into this: this PR archives symlinks as symlinks, and does not follow them. (Right?)

zip.go
```diff
@@ -186,6 +186,15 @@ func (z *Zip) extractNext(to string) error {
 	if !ok {
 		return fmt.Errorf("expected header to be zip.FileHeader but was %T", f.Header)
 	}
+	if (header.FileInfo().Mode() & os.ModeSymlink) != 0 {
+		buffer := make([]byte, header.FileInfo().Size())
```
mholt: This seems to be very different from how it works in tar archives; while I know that this PR writes the target filepath as the symlink's contents, is it safe to assume that every zip archiver will do that? Is there risk here of reading in a potentially huge file?

zip.go
```diff
@@ -317,6 +328,18 @@ func (z *Zip) Write(f File) error {
 		return nil
 	}
 
+	if (header.Mode() & os.ModeSymlink) != 0 {
```
mholt: What do you think about mirroring the more elegant switch over in the tar extractor? (perhaps sans hard link of course):

https://sourcegraph.com/github.com/mholt/archiver@d572b2e8b82726cee9476d1b9d63a7fe9b601ff1/-/blob/tar.go#L232

zip.go
jandubois: I'm not 100% sure; I simply went by the symlink code in archive/zip test: https://github.com/golang/go/blob/b0a53d2/src/archive/zip/writer_test.go#L54-L59

I believe this is what Info-ZIP refers to as the "generic symlink support" in the later versions. Earlier versions (and PKZIP) had platform specific extensions for symlinks (so symlinks archived on one platform could not be extracted on another one), but I think those have been abandoned.

I'm not sure if this is necessary, but I can add a check for some reasonable maximum path length, if you prefer. If you want that, what should the limit be? `2^16` seems to be way beyond what I would expect to see for real pathnames, but small enough not to cause memory issues...

mholt: Hmm, I suppose this is OK then, but I would probably want to move this logic to the `z.extractFile` method, if possible.

zip.go
jandubois: I don't understand how this should work. The TAR `Header` has an explicit `Typeflag` byte, that you can switch on:

https://github.com/golang/go/blob/ea509c9/src/archive/tar/common.go#L144

The ZIP `FileHeader` has a `Mode()` function that needs to locate the mode bits depending on the "creator" version:

https://github.com/golang/go/blob/ea509c9/src/archive/zip/struct.go#L277-L289

So we would have to switch on `header.Mode() & os.ModeType` and could then check for `os.ModeSymLink`, and an explicit `0` for `IsRegular()`? Essentially:

```
	switch header.Mode() & os.ModeType {
	case os.ModeSymlink:
		linkTarget, err := os.Readlink(f.Name())
		if err != nil {
			return fmt.Errorf("%s: readlink: %v", f.Name(), err)
		}
		_, err = writer.Write([]byte(filepath.ToSlash(linkTarget)))
		if err != nil {
			return fmt.Errorf("%s: writing symlink target: %v", f.Name(), err)
		}
	case 0: // regular file
		if f.ReadCloser == nil {
			return fmt.Errorf("%s: no way to read file contents", f.Name())
		}
		_, err := io.Copy(writer, f)
		if err != nil {
			return fmt.Errorf("%s: copying contents: %v", f.Name(), err)
		}
	}
	return nil
}
```

I can do this if you prefer; let me know.

mholt: Oh, I see what you mean now.

Hmm, maybe a switch isn't quite right then, you've got a point. An `if` statement could do. Then perhaps this should be brought out into its own method, like `z.writeFile`. That'll satisfy my OCD on this. :smile:

zip.go
jandubois: I put the logic into `extractNext` because it already had done the type assertion on `header`, but you are right, it must be in `extractFile`, otherwise `Extract` will not be able to extract symlinks.  I'll do this (and duplicate the `header` assertion into `extractFile`).

Unfortunately this points out that test coverage of the `Extract` method is lacking. 😦

zip.go
jandubois: I don't know; moving it into a separate function will require duplication of the `header` assertion logic, so if there is only a single caller, I would prefer to avoid the code duplication.

I've pushed the changes I have so far, as I'm done for today. I can always rewrite it some more if you don't like the switch either...

jandubois: > That'll satisfy my OCD on this.

Talking about OCD, it took me a lot of willpower not to change `hdr` to `header` in the rest of the codebase to get consistent naming everywhere. 😄

tar.go
```diff
@@ -338,7 +340,16 @@ func (t *Tar) Write(f File) error {
 		return fmt.Errorf("missing file name")
 	}
 
-	hdr, err := tar.FileInfoHeader(f, f.Name())
+	filename := f.Name()
```
mholt: FYI, previous commits in this change had left the `filename` (previously called `linkTarget`) empty for non-symlinks, which I _think_ was probably a bug. Should be fixed now.

mholt:
Thanks for the work! I took the liberty of finishing up a few minor changes. Tests are passing; let me know what you think, and if it works for you on your symlink-containing archives.

zip.go
mholt: Heh, you shouldn't have told me about that 😣

tar.go
mholt: Oh, you're right. Pushed a commit that fixes that. Thanks. Am ready to merge if you are!

jandubois: > Am ready to merge if you are!

Sure, go ahead and let's close a bunch of open issues! 😄 

I assume you will take care of filing issues for things discovered in this PR, if you feel they should be addressed at some point:

* Add option to dereference symlinks before adding to archive

* Add test coverage for `Extract` method

* (maybe) Remove vestigial hardlink support

* (maybe) Investigate weird permissions issue with ZIP files created on Windows

mholt: Yep, I think those are all suitable for other issues and PRs. Thanks again!

PR #98, opened 2018-11-05
Update zip.go

lucasxu00:
Remove compressed formats check to avoid zip header check exception in java.util.zip.ZipInputStream(only DEFLATED entries can have EXT descriptor)


mholt:
Sorry, but this seems like a Java problem, not an issue with archiver, which is fully spec-compliant in this matter...

PR #99, opened 2018-11-06
Rewrite entire package

### LINKED ISSUE #22: Support for avoiding overwrite upon opening archive

### LINKED ISSUE #39: Is there support for password protected archives?

### LINKED ISSUE #58: Add possiblity to only extract some files of an archive
Hi,

I need to only extract 1 file (no more that 10KB, that I made it be first in the archive) of a very big .tar archive (more than 1GB).

Could you add a way to provide a list of file that we want to extract without extracting every files.

Thanks.

### LINKED ISSUE #61: Zip output path set to non-existing directory casues failure
This may be able to be a non-issue, but I was expecting that when I passed a directory location that did not yet exist, the directory would get created and the zip would still get output to that path.  

    err := archiver.Zip.Make(outputPath, files)

Instead, it returns an error:

> error creating [outputPath]: open [outputPath]: The system cannot find the path specified.

I created the output directory structure and the previous code worked fine.  

When zipping using a desktop application, you can provide an nonexisting output path and the directories will be created for you.  So, who should be the owner of creating the directories?

I'm on Win10 x64

### LINKED ISSUE #68: Support for Multi-part Archives
I'm attempting to extract rar files that are multi-part (.rar .r01 r.02 etc). Using the `unrar` tools available out there, I just pass them a single file name (the .rar file) and they extract all the subsequent archives. I'm using `archiver` as a library and calling `archiver.Rar.Open(file, path)` then receive this error: `writing file: rardecode: archive continues in next volume`

Am I missing something, or is this support missing? Thanks!

### LINKED ISSUE #80: Feature request: (de)compression only (single-file archives)
I've been using archiver for a while for a [tool](https://github.com/holen-app/holen) that downloads releases from github (and other places) and automatically unpacks them to make their commands available for running.  Archiver has worked out very well so far.

Today, I wanted to add [restic](https://restic.net/) to my tool, and it's [downloads](https://github.com/restic/restic/releases/tag/v0.9.2) are only bzip2 compressed, as opposed to being tar'd first.  This is not supported by archiver.

It looks like it would be possible to add support for archives of one file that are only compressed (gz or bzip2).  I am wondering, would this support be in line with the goals of the project and likely be accepted?  I want to know before investing more effort.

Thanks for making such a useful library.

### LINKED ISSUE #90: Proposal: Rewrite the entire package. Design doc discussion
## UPDATE: Done in pull request #99

Having been using this package for a few years now, I've encountered a number of issues that lead me to want to redesign this package entirely: burn it down and start over, copying only the fundamental parts of the code, and not worrying about backwards compatibility.

Some specific issues I've experienced:

- **Too much magic.** Recently I spent a day debugging an problem where a .zip file couldn't be extracted every time with archiver. Sometimes it would work, sometimes it wouldn't. I eventually discovered that it's because archiver determines which extractor to use based on the extension and the file header while iterating through a map of formats (which is not ordered). If the Zip format came first, it matched by extension but failed to extract; if the TarGz format came first, it matched by file header (because it was actually a tar.gz file), and extraction succeeded.

- **Weak API.** Apparently I was able to accidentally create a .tar.gz file with the Zip archiver, because the name I built for the file was not attached to which archiver format I was using. I can do `archiver.TarGz.Make("file.zip")` without errors, which is bad. Here's the code that led to my bug in the first place (notice the missing `.` in `"zip"`):

```go
var a archiver.Archiver = archiver.TarGz
if filepath.Ext(outputFile) == "zip" {
	a = archiver.Zip
}
```

^ Bad package design.

- **Not enough customizability.** Namely: compression level; whether to include a top-level named folder vs. just its contents), similar to how rsync works based on presence of a trailing slash; and whether to overwrite existing files when outputting.

- **Lack of native streaming capabilities.** Recently there were From a library perspective, I should be able to stream in a zip file and spit out individual files, or stream in individual files (or a list of filenames?) and spit out a single zip file.

- **[There is no true cross-platform native solution to zip-slip (yet)](https://github.com/rubyzip/rubyzip/pull/376).** I had to disable the "security feature" that prevented me extracting a perfectly safe archive. Even "SecureJoin" solutions don't cut it (read the linked thread, and its linked threads). For now, these "mitigations" only get in the way.

- **Not enough power to inspect archives or cherry-pick files.** It would be helpful to be able to work with archives' contents without performing an extraction, such as getting listings, or filtering which files are extracted, etc.

General solutions:

- When possible (almost always), match only by file header and ignore file extension. If the file contents are not (yet) available, then use extension but only after a warning or explicit opt-in. Or, (maybe "Also,") require that the file extension, where present, matches the format when creating an archive.

- Be verbose in the error messages; if doing any magic, report it or make the magic explicitly opt-in either with a configuration parameter or a special high-level API that is documented as being magical, which wraps the underlying, concrete, explicit functions.

- Couple the file extension to the archiver. For example: don't allow the Zip archiver to make a .tar.gz file. For example, the buggy code above could have been avoided with something more like this: `archiver.Make(outputFile, files...)` which uses the extension of `outputFile` to force a format that matches.

- Expand the API so that an archiver is created for a specific format before being used, rather than having hard-coded globals like archiver.Zip like we do now. This will allow more customization too. Imagine `zipArch := archiver.Zip{CompressionLevel: 10}` or something similar.

- Be explicit about our threat model, which is being adjusted, to state that the files are expected to be trusted, i.e. don't download files you don't trust. Maybe it is possible to inspect a file before extracting it to know whether it could be malicious (e.g. look for zip-slip patterns in file names), but I am not sure about that yet.

- Moar interfaces. We have one, Archiver, but we might need more, to accommodate an expanded design with more features. Small interfaces are the best.

- ~~**Rename the package to `archive`**.~~ (Decided to keep it the same)

This issue is to track the discussion about the new design; work will hopefully begin soon, as I can find the time.

### LINKED ISSUE #97: Slow extract for tar.xz
https://github.com/ulikunitz/xz is very slow at extracting. 

https://github.com/xi2/xz is 10x faster for me. Consider using this for extraction. It doesn't support compression, so it won't be a full replace.

mholt:
Alright, this little library has become quite an important tool for me over the last couple years and it's time to give it some TLC.

For background, see #90 (and frankly all other open and even closed issues).

**This change supersedes all open PRs.**  Any open PRs will need to be reconsidered and, if still relevant, rebased. 

Closes #90. <-- main issue
Closes #22.
Closes #39.
Closes #58.
Closes #61.
Closes #68.
Closes #80.
Closes #97.

## Format progress

- [x] .zip
- [x] .tar
- [x] .gz
- [x] .bz2
- [x] .xz
- [x] .lz4
- [x] .sz
- [x] .tar.gz
- [x] .tar.bz2
- [x] .tar.xz
- [x] .tar.lz4
- [x] .tar.sz
- [x] .rar (extract-only)

## Other features progress

- [x] CLI
- [x] Port existing tests

## Highlights

- Archivers are created like this (most fields optional - there's also `DefaultZip`, etc, with some sane defaults that are ready to use):

```go
z := archive.Zip{
	CompressionLevel:       flate.DefaultCompression,
	MkdirAll:               true,
	SelectiveCompression:   true,
	ContinueOnError:        false,
	OverwriteExisting:      false,
	ImplicitTopLevelFolder: true,
}
```

- They can create an archive like this:
```go
err := z.Archive([]string{"file1.txt", "../file2.txt"}, "test.zip")
```

- When creating archives, file extension MUST match the archive type.

- They can open and extract whole archives like this:
```go
err = z.Unarchive("test.zip", "/Users/matt/Desktop")
```

- They can traverse/walk/inspect archives without extracting them:
```go
err = z.Walk("test.zip", func(f archive.File) error {
	zfh, ok := f.Header.(zip.FileHeader)
	if ok {
		fmt.Println(zfh.Name)
	}
	return nil
})
```

- Inside a walk function, you can even read the contents of files if you want (yep, that's right, they're `io.ReadCloser`s! - they get closed for you when the walk function returns):
```go
_, err = io.Copy(w, f)
```

- They can extract single files (or folders, recursively) from an archive:
```go
err = z.Extract("test.zip", "testdata/myfile.txt", "/Users/matt/Desktop")
```

- They can even read and write archives in a streaming fashion, file-by-file, to any writer or from any reader:
```go
// STREAMING WRITE EXAMPLE:
err = z.Create(out)
if err != nil {
	return err
}
defer z.Close()
// ... open a file, or any io.ReadCloser ...
err = z.Write(File{
	FileInfo: archive.FileInfo{
		FileInfo:   file.Info(),
		CustomName: "name/in/archive.txt",
	},
	ReadCloser: file,
})

// STREAMING READ EXAMPLE:
// (can be any io.Reader in general, but zip requires io.ReaderAt and knowing the size)
err = z.Open(file, fileInfo.Size())
if err != nil {
	return err
}
defer z.Close()

for {
	f, err := z.Read()
	if err == io.EOF {
		break
	}
	if err != nil {
		return err
	}
	// you can access the header now through f.Header or
	// read the contents as f is also an io.ReadCloser, then
	// close f when finished
	err = f.Close()
}
```

It should work similarly for the other archive formats.

Single-file compression and decompression is possible, too.

This PR includes a fully functional CLI. One feature request I would still add is the ability to stream compressors and decompressors through the stdin and stdout, but that can come later. Very easy to add.

We can potentially implement options to configure symlink handling and even scan for zip-slip patterns and prevent an ugly extraction (but we cannot fix zip-slip implicitly, unfortunately).


archive/tar.go
```diff
@@ -0,0 +1,527 @@
...
+			// finished walking the entire directory
+			return ErrStopWalk
+		}
+
+		return nil
+	})
+}
+
+// Compile-time checks to ensure type implements desired interfaces.
+var (
```
komuw: I learn something new everyday. 
thanks for this.

PR #114, opened 2018-11-16
Implement CheckExt(string) error for Archiver interface

frebib:
There is no functional difference for existing code, only the additional function `Archiver.CheckExt(string) error` which can be used to validate file extensions for Archiver interface types.

As for stylistic choices, I wasn't too sure whether I should return the error immediately, wrap it (like I have done) or create a new error. The implementation I chose retains the original functionality whilst also matching the `CheckExt(string) error` return values in the `Compressor` interface. It's fortunate that the error strings are formatted to let me do it like this.

Also I was worried about a name-clash causing confusion/compile errors with the `Compressor.CheckExt(string) error` function, so let me know if you think I should change it.

### DIFF
--- a/archiver.go
+++ b/archiver.go
@@ -13,12 +13,18 @@ import (
 // Archiver is a type that can create an archive file
 // from a list of source file names.
 type Archiver interface {
+	ExtensionChecker
+
 	// Archive adds all the files or folders in sources
 	// to an archive to be created at destination. Files
 	// are added to the root of the archive, and directories
 	// are walked and recursively added, preserving folder
 	// structure.
 	Archive(sources []string, destination string) error
+}
+
+// ExtensionChecker validates file extensions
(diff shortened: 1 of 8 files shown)


archiver.go
```diff
@@ -19,6 +19,7 @@ type Archiver interface {
 	// are walked and recursively added, preserving folder
 	// structure.
 	Archive(sources []string, destination string) error
+	CheckExt(name string) error
```
mholt: This might sound pedantic, but how about a new interface called `ExtChecker` or `ExtensionChecker` with this method in it, then just embed that interface into Archiver and Compressor.

frebib: Gotcha, makes sense!

frebib:
I'm not sure how you mark the changes as 'fixed' but either way I've made the changes

mholt:
Thank you!

PR #120, opened 2018-11-20
Add and use ErrFormatNotRecognized (fixes #117)

### LINKED ISSUE #117: Make "format not recognized" error a const
https://github.com/mholt/archiver/commit/e1ccfb528128bf4ff3bd906823ae2dcb708f79f5#diff-635e4219ee55ef011b2b32bba065606bR405

That way we can do `if err != nil && err != archiver.ErrNotRecognized`.

onodera-punpun:


archiver.go
```diff
@@ -390,7 +394,7 @@ func ByExtension(filename string) (interface{}, error) {
 	case *Xz:
 		return NewXz(), nil
 	}
-	return nil, fmt.Errorf("format unrecognized by filename: %s", filename)
+	return nil, ErrFormatNotRecognized
```
mholt: This is a different error, even though the wording is similar, and shouldn't be changed.

archiver.go
```diff
@@ -126,6 +126,10 @@ type WalkFunc func(f File) error
 // ErrStopWalk signals Walk to break without error.
 var ErrStopWalk = fmt.Errorf("walk stopped")
 
+// ErrFormatNotRecognized is an error that will be send if
```
mholt: "will be returned"

PR #121, opened 2018-11-20
Add method parameter to Unarchive function (fixes #118)

### LINKED ISSUE #118: Add ByHeader support to all formats
Or at least, make it use this by default. The reason being that `ByExtension()` supports many more formats.

onodera-punpun:
I don't know it using a string for this is the best option here?

### DIFF
--- a/archiver.go
+++ b/archiver.go
@@ -161,18 +161,27 @@ func Archive(sources []string, destination string) error {
 }
 
 // Unarchive unarchives the given archive file into the destination folder.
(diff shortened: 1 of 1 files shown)


mholt:
No... sorry... this is not the right way to go about this. I still think the right way to do it is by the header.

mholt:
To clarify what I mean, let's make ByHeader support as many formats as ByExtension does. Would be happy to reopen it if we do that.

Issue #131, opened 2018-12-18
Why did Compress and Uncompress disappear?


simonjohansson:
Howdy and thanks for a awesome library!

We have a use case where we tar up a path and upload it to google cloud storage, which takes a []byte as the object you wish to upload. Later on we might download the tar which again gives us a []byte and untar it to some path.

In `v2.0.1` we have the super helpful

```
Compress(sourceDir string) (tar []byte, err error)
Uncompress(destinationDir string, tar []byte) (err error)
```

We noticed that we had issues with the symlink and that later versions of archiver fixes this. But the helpful functions are now gone in favor of 

```
Archive(sources []string, destination string) error
Unarchive(source, destination string) error
```

Any reason why there are no methods do archive into []byte and unarchive from []byte anymore?

mholt:
Hmm, I can't find those functions in the commit history -- do you have a link to them to refresh my memory?

In any case, I think it must have been removed because the use case for buffering entire archive files in memory as byte slices was limited. It's much more recommended to stream instead. (How would you upload large objects to the cloud? That API sounds terrible...) There are examples how to do this in the godocs: https://godoc.org/github.com/mholt/archiver#example-Zip--StreamingWrite

mholt:
Hm, okay -- I'll tag this as a feature request; contributions welcomed.

hoster110:
My idea is to use full IO to accelerate the compression efficiency;
1. Can multi-threading be supported?
2. Can you write zip file in an additional way to adapt to the memory shortage caused by large files?

hoster110:
func Zip(srcFile string, destZip string, thNums string) error {
	SytemToInfo()
	err := filepath.Walk(srcFile,Listfunc)
	if err != nil {
		Err(err)
		return err
	}
	nums, err := strconv.ParseInt(thNums,10,32)
	if err != nil {
		Err(err)
		return err
	}

	batchsize := 2

	for index := 0; index < len(ListFile); index = index+batchsize{

		wg.Add(1)
		go func (i int){
			buf := new(bytes.Buffer)
			archive := zip.NewWriter(buf)

			defer archive.Close()
			for in := i;in < i+batchsize && in < len(ListFile);in++{
				ff, err := archive.Create(ListFile[in])
				if err != nil {
					Err(err)
					continue
				}

				bytes,err := ioutil.ReadFile(ListFile[in])
				_, err = ff.Write(bytes)
				if err != nil {
					Err(err)
					continue
				}

			}
[...]

PR #132, opened 2018-12-21
Spikey implementation of archiving into and unarchiving from streams

simonjohansson:
Merry Christmas!

Following https://github.com/mholt/archiver/issues/131 here is a spikey implementation of bringing back the functionality.

If you agree on the general implementation Ill clean it up, add more tests and implement it for the other archive types.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ _gitignore
 builds/
 *.test
 cmd/archiver/archiver
+.idea
--- a/archiver.go
+++ b/archiver.go
@@ -65,6 +65,7 @@ type Archiver interface {
 	// are walked and recursively added, preserving folder
 	// structure.
 	Archive(sources []string, destination string) error
+	ArchiveToStream(output io.Writer, sources []string) error
 }
 
 // ExtensionChecker validates file extensions
@@ -76,6 +77,7 @@ type ExtensionChecker interface {
 // into a folder.
 type Unarchiver interface {
 	Unarchive(source, destination string) error
+	UnarchiveFromStream(input io.Reader, destination string) error
 }
 
 // Writer can write discrete byte streams of files to
--- a/tar.go
+++ b/tar.go
@@ -51,6 +51,8 @@ type Tar struct {
 	readerWrapFn  func(io.Reader) (io.Reader, error)
 	writerWrapFn  func(io.Writer) (io.Writer, error)
 	cleanupWrapFn func()
+
+	destination string
 }
 
 // CheckExt ensures the file extension matches the format.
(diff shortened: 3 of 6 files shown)


.gitignore
```diff
@@ -3,3 +3,4 @@ _gitignore
 builds/
 *.test
 cmd/archiver/archiver
+.idea
```
mholt: Is this an IntelliJ folder?

archiver.go
```diff
@@ -65,6 +65,7 @@ type Archiver interface {
 	// are walked and recursively added, preserving folder
 	// structure.
 	Archive(sources []string, destination string) error
+	ArchiveToStream(output io.Writer, sources []string) error
```
mholt: Instead of modifying existing interfaces, let's make new ones: `StreamArchiver` and `StreamUnarchiver`, maybe. And, although this is yak-shaving, let's reverse the order of the arguments to match Archive and Unarchive; in, out; not out, in.

tar.go
```diff
@@ -51,6 +51,8 @@ type Tar struct {
 	readerWrapFn  func(io.Reader) (io.Reader, error)
 	writerWrapFn  func(io.Writer) (io.Writer, error)
 	cleanupWrapFn func()
+
+	destination string
```
mholt: I bet we can find a way to implement this without adding more state.

mholt:
Oh, oh! I have a better idea than what I just suggested: how about two new package-level functions, instead of new interfaces:

```go
func ArchiveToStream(sourceFiles []string, out io.Writer, a Archiver) error

func UnarchiveFromStream(in io.Reader, destination string, u Unarchiver) error
```

What do you think of this?

.gitignore
simonjohansson: Yah

tar.go
simonjohansson: Yeah, for sure. I just made it in the most bruteforce way possible without having to pass around empty strings to methods

simonjohansson:
i quite like the idea of the interfaces, as I think it would be quite neat to have the functionality available for all types of archives :)

mholt:
@simonjohansson That's still possible with the function approach -- the third argument is any archiver.Reader/Writer, which both Tar and Zip implement.

mholt:
Also I think with the function approach (instead of making new interfaces), it'll "just work" for any archive format that already implements Reader and Writer interfaces, meaning you don't have to modify the Zip and Tar types at all.

simonjohansson:
Ah right, now I get it :)

Ill continue the work after the holidays!

argos83:
Hi! is there any plan to resuscitate this PR? I would like an easy way to archive a directory recursively into a writer instead of a file. My use case is that I want to send a tar.gz of a directory in a request body, something like this:

```go
reader, writer := io.Pipe()
tarGz := archiver.NewTarGz()
tarGz.ArchiveToStream([]string{"./some-dir/."}, writer)

req, _ := http.NewRequest(http.MethodPost, "http://localhost/some-path", reader)
... send request..
```
I just found about this library, and from what I read in the docs I believe my options currently are:

 * use `tarGz.Create(writer)`, walk the directory myself, open every file and do `tarGz.Write(file)`
 * or use `tarGz.Archive` to save the file somewhere in the filesystem, and then read it to send the request.

The second option seams simpler to me, but I would prefer not having to write in the file system (nor having to clean up after). The changes in this PR seems to be what I'm looking for, or is there another alternative I'm missing?

Thanks!

mholt:
@argos83 There is an example of how to compress directly to a stream in both the readme and the godocs: https://godoc.org/github.com/mholt/archiver#example-Zip--StreamingWrite

mholt:
Thanks so much for contributing this, but this will be superceded by #302, which is a complete rewrite of archiver with new stream-oriented APIs. Please let me know if there's any changes or improvements I should make! :)

PR #135, opened 2019-01-03
Fixed Rar directory extracting

AzazKamaz:
When extracting directory, library extracts it as file (trying to create (directory may exist if a file contained in has been extracted early) and cause an error about creating file on catalog) (or create it as file and break next files extraction (if them are contained in that folder)


mholt:
Thanks -- can you please add a test case to verify the behavior?

archiver_test.go
```diff
@@ -251,6 +251,26 @@ func TestArchiveUnarchive(t *testing.T) {
...
+
+		dest := filepath.Join(tmp, "extraction_test_"+auStr)
+		os.Mkdir(dest, 0755)
+
+		file := "testdata/testdata.rar"
+		err = au.Unarchive(file, dest)
+		if err != nil {
+			t.Fatalf("[%s] extracting archive [%s -> %s]: didn't expect an error, but got: %v", auStr, file, dest, err)
+		}
+	}
```
mholt: Does this test actually confirm that an empty folder in a rar file is created when extracting? It looks like it just checks to make sure the `Unarchive` function doesn't return an error.

AzazKamaz: But it checks that my fix is working. Without it, there was an error when extracting

mholt: Ah, I see. I guess that'll do then.

mholt:
@AzazKamaz I had to move testdata.rar into the testdata folder, and the sample files into a corpus subfolder, in 2bd00e47f6e10b3199ae839ca3ccaea9c9f3aa32. I don't have a way of making rar files, any chance you could update the sample.rar file so the test I had to comment out will pass again? :)

PR #138, opened 2019-01-09
Add support for byheader format check for tar.gz files

johnarok:
Supports #138


mholt:
Thank you for this. Sorry I never got around to this. It'll be superseded by #302 which is a complete rewrite of this package, and will become v4. It supports identifying files by their headers, for all formats including tar and any compression format in combination. Thanks again!

PR #139, opened 2019-01-15
remove probable existing symbolic link

meysampg:
On some extracting situations that an archive contains a symbolic link file,
archiver tries to create a symbolic link, but already the symbolic file
exists and it causes the `file exists` error. This commit adds a blind
removing try: if the symbolic link already exists (from extraction process),
remove it and if does not, ignore the returned error.

### DIFF
--- a/archiver.go
+++ b/archiver.go
@@ -305,10 +305,13 @@ func writeNewFile(fpath string, in io.Reader, fm os.FileMode) error {
 	return nil
 }
 
-func writeNewSymbolicLink(fpath string, target string) error {
+func writeNewSymbolicLink(fpath string, target string, overwriteExisting bool) error {
 	err := os.MkdirAll(filepath.Dir(fpath), 0755)
 	if err != nil {
 		return fmt.Errorf("%s: making directory for file: %v", fpath, err)
+	}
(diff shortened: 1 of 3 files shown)


archiver.go
```diff
@@ -310,6 +310,7 @@ func writeNewSymbolicLink(fpath string, target string) error {
 	if err != nil {
 		return fmt.Errorf("%s: making directory for file: %v", fpath, err)
 	}
+	os.Remove(fpath)
```
mholt: We should only remove it if it's a symbolic link, right? Not safe to overwrite existing files with actual content.

Also might need to honor the various OverwriteExisting parameters for each archiver, for example, zip's: https://github.com/mholt/archiver/blob/d8fc61af55e91ce2231ac9e1965cb9e77b435be8/zip.go#L25

meysampg: @mholt Please consider new changes. 
Now `zip` and `tar` archives which currently use `writeNewSymbolicLink` method, use their `OverwriteExisting` to remove existing symbolic links safely.

mholt:
I believe #302 makes this irrelevant -- thank you for the contribution. Sorry I never got around to finishing it with you.

PR #142, opened 2019-01-31
bugfix extract single file from zip

hy05190134:


mholt:
Thank you for the PR, but can you explain this? And give a reproducible case, and provide a test?

hy05190134:
end, err := filepath.Rel(targetDirPath, zfh.Name)		
 			if err != nil {		
 				return fmt.Errorf("relativizing paths: %v", err)		
 			}		
 			joined := filepath.Join(destination, end)

when exec the code above, and when zfh.Name == "abc/def.doc" and destination == "./music", then joined will return "./music/abc/def.doc", but the function ExtractSingleFile will append "abc/def.doc" again, so the real path is "./music/abc/def.doc/abc/def.doc", you can test for single

discreet:
@mholt @hy05190134 

This is the exact functionality I need. My use case is I have a tarball that contains a binary and I want to extract that binary directly to `/usr/local/bin`. Since the binary is not in the root of the tarball instead of getting `/usr/local/bin/<binary>` I get `/usr/local/bin/<parent dir>/<binary>`.

I know there is a risk here of file collision if you have the same named file under two different directories in the archive but maybe something can be implemented to optionally drop the parent directory?

hy05190134:
@discreet  so just OK, I modify the source code by myself to fit my case.

mholt:
I believe with #302 this is irrelevant now, as you have more control over file extraction.

Issue #143, opened 2019-02-07
Unlink the symbolic link if it already exists


jchionh:
The function `writeNewFile()` overwrites the file if it already exists. Following the same behavior, should `writenewSymbolicLink()` also overwrite the symbolic link if it already exists?

If this is agreeable, I propose to do so by checking if the symlink exists, and unlink using `os.Remove()` if true, before doing `os.SymLink()`

jchionh:
Referencing this function:
https://github.com/mholt/archiver/blob/master/archiver.go#L313

mholt:
Hm, yes, that seems like a good idea; they are used in parallel just depending on the file mode.

jchionh:
https://github.com/mholt/archiver/pull/145

jchionh:
PR #145 was merged, closing this issue.

PR #144, opened 2019-02-07
Unlink symbolic and hard links if they already exists while extracting

jchionh:
The function `writeNewFile()` overwrites the file if it already exists.
Following the same behavior, modified `writeNewSymbolicLink()` and
`writeNewHardLink()` to do the same.

Referencing this issue:
https://github.com/mholt/archiver/issues/143


jchionh:
fmt errors, will fix and do a separate PR

Issue #165, opened 2019-04-29
[ERROR] Reading file in tar archive: foobar/foobar.qcow2: unknown type flag: S


praveenkumar:
## What version of the package or command are you using?
```console
$ arc unarchive crc_libvirt_v4.1.0.rc0.tar.xz 
```

## What are you trying to do?
Trying to extract a `xz` tarball which created using `tar cJSf`.

```console
$ ls -lh crc_virtualbox_v4.1.0.rc0
total 20440536
-rw-r--r--  1 prkumar  staff   120B Apr 29 10:17 Corefile
-rwxr-xr-x  1 prkumar  staff    46M Apr 29 10:17 coredns
-rw-r--r--  1 prkumar  staff   722B Apr 29 10:01 crc-bundle-info.json
-rw-r--r--  1 prkumar  staff   9.7G Apr 29 09:59 crc.vmdk
-rwxr-xr-x  1 prkumar  staff   5.7K Apr 29 10:20 crc_virtualbox.sh
-r--------  1 prkumar  staff   1.8K Apr 28 10:17 id_rsa_crc
-rw-r--r--  1 prkumar  staff    23B Apr 28 10:17 kubeadmin-password
-rw-r--r--  1 prkumar  staff   8.7K Apr 28 10:17 kubeconfig
-rw-r--r--  1 prkumar  staff   431B Apr 29 10:18 test1-api
```

## What did you expect to happen, and what actually happened instead?
I would expect it should've extracted all the files from tarball but got an folder which doesn't contain the `vmdk` file.

```console
$ ./arc unarchive crc_libvirt_v4.1.0.rc0.tar.xz 
2019/04/29 15:50:30 [ERROR] Reading file in tar archive: crc_libvirt_v4.1.0.rc0/crc.qcow2: unknown type flag: S
```

## How do you think this should be fixed?
I am thinking this might be happening because of the file size is too big.

## Bonus: What do you use archiver for, and do you find it useful?
we want to use it for a project https://github.com/code-ready/crc/blob/master/pkg/crc/machine/bundle/extract.go as of now using gzip tarball but want to switch to `xz` one.

PR #166, opened 2019-04-29
Add TypeGNUSparse as tar type to handle vmdk,qcow2 images

praveenkumar:
Fix issue #165 

Add the go modules

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,10 @@
+module github.com/mholt/archiver
+
+require (
+	github.com/dsnet/compress v0.0.1
+	github.com/golang/snappy v0.0.1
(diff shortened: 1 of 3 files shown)


gbraad:
Works for me

gbraad:
Note: `GO111MODULE=auto` is set and fails the build

cfergeau:
This really is  about the use of -S/--sparse when generating the tarball (and adding a sparse file to it), this not related to vmdk, qcow2 or gz vs xz (I was initially misled by the commit logs :)

mholt:
Closing due to inactivity; will add the flag in a separate commit.

PR #179, opened 2019-08-25
Add KeepParentDirectories flag 

chriswalz:
**When zipping something like this the paths were mapping like:**

web/static -> static
configs/os_agnostic/deployment -> deployment

**I want something like this**

web/static -> web/static
configs/os_agnostic/deployment -> configs/os_agnostic/deployment

### DIFF
--- a/zip.go
+++ b/zip.go
@@ -50,6 +50,11 @@ type Zip struct {
 	// especially on extraction.
 	ImplicitTopLevelFolder bool
 
+	// When zipping parent/folderName
+	// the zip file will contain parent/folderName
+	// as opposed to just folderName
+	KeepParentDirectories bool
+
 	// If true, errors encountered during reading
 	// or writing a single file will be logged and
 	// the operation will continue on remaining files.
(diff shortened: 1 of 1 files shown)


chriswalz:
Any thoughts on this?

mholt:
Can you be more specific about what you are doing, what the actual behavior is, and what the expected behavior is?

chriswalz:
**Background**

I wrote a go server that has html files stored in a folder `web/html`. The server uses this path to find the html files that it serves. 

When I zip these files and copy them to a linux server in the cloud and unzip them the path is now just `html` (The parent folder `web` no longer exists). Now when my server tries to serve files from `web/html` it thinks they don't exist since they're really in `html` folder. 

I hope that's more clear. 

**Actual Behavior**

web/html -> html

**Expected (with `KeepParentDirectories` flag)** 

web/html -> web/html

chriswalz:
Also, note I originally had a bash script that was using mac terminal `zip` and `unzip`. 

When I converted this script to Go, using `Archiver` package it wasn't working anymore since `zip`'s output is different than `Archiver`'s with the same parameters hence why I'm suggesting adding this flag. 

Perhaps I'm missing something and one of the other flags do this but it doesn't appear to be so.

mholt:
I _almost_ understand, but if you could provide a few lines of code (or the exact commands you're using, if the CLI) to demonstrate what you mean, that would be best.

chriswalz:
Expected behavior using (`zip` and `unzip`)
expected

Actual behavior using (`arc archive` and `unzip`)
actual

mholt:
Awesome, thank you!

chriswalz:
So can we merge it in? What's the next step

mholt:
I just need to find some time to review it. Sorry, it's on me.

mholt:
v4, introduced in #302, gives you this flexibility now, so I think this is resolved/irrelevant in the new version.

Issue #150, opened 2019-03-04
Method to extract zip archive from memory


farshidtz:
Hi,

In the version 2 of the package, one can simply extract zip archives from memory using the [Zip.Read](https://github.com/mholt/archiver/blob/v2.1.0/zip.go#L153) method:
```
archiver.Zip.Read(bytes.NewBuffer(zippedBytes), "dest/path")
```

It would be nice to add a similar method to version 3. I managed to achieve this with 50 lines of code (Zip.Open and then Zip.Read + file writes in a loop) but couldn't maintain the file hierarchy. This is doable using Go's `archive/zip`.

Thanks

gadelkareem:
I was looking for a similar function to read from bytes and returns a reader just like the zip lib `zip.NewReader(bytes.NewReader(ib), int64(len(ib)))`
It would be great if we can totally avoid disk writes ex: 
`archiver.NewReader(bytes.NewReader(ib), int64(len(ib))), "filename.ext")`

1138-4EB:
Same issue here. With `archive/zip` I can replace `OpenReader` with `NewReader`:

```go
	r, err := zip.OpenReader(src)
	if err != nil {
		return nil, err
	}
	defer r.Close()
```

```go
	body, err := ioutil.ReadAll(src)
	if err != nil {
		return nil, err
	}

	r, err := zip.NewReader(bytes.NewReader(body), int64(len(body)))
	if err != nil {
		return nil, err
	}
```

But this seems not to be possible with `archiver`. @mholt is this a `wontfix`?

nmiyake:
Will note that there is some similarity here to #131 , which is asking for APIs to archive and unarchive from `io.Reader` and `io.Writer` interfaces.

J7mbo:
This looks promising - any progress?

ViRb3:
I achieved universal extraction from memory by downcasting `archiver.NewZip()` or `archiver.NewTarGz()` to `archiver.Reader`, and then:
```go
for {
		file, err := reader.Read()
		if err == io.EOF {
			break
		} else if err != nil {
			return err
		}
		filePath := filepath.Join(basePath, file.Name())
		if file.IsDir() {
			if err := os.MkdirAll(filePath, 755); err != nil {
				return err
			}
		} else {
			data, err := ioutil.ReadAll(file)
			if err != nil {
				return err
			}
			if err := ioutil.WriteFile(filePath, data, 644); err != nil {
				return err
			}
		}
	}
```

[...]

PR #182, opened 2019-09-26
add UnarchiverIO

umarcor:
This is a proof of concept to allow unarchiving files from memory, without saving them to disk.

Ref #150


mholt:
Thank you for the contribution! This PR will be superseded by #302, where I completely rewrote the archiver package in preparation for v4. Its new core APIs are stream-oriented, so you'll be able to operate completely in memory and with way more flexibility.

PR #185, opened 2019-10-18
Improved Close() error handling, updated zstd library, and zstd codec options exposed.

### LINKED ISSUE #184: Close errors silently ignored
In latest `master`, the signature of `cleanupFunc` is `func()`, silences errors. Its use through is mostly through `defer` which also silently ignore errors.

All of the `.Close()` funcs silently ignore the error (because `cleanupFunc` doesn't handle it), therefore, if for some reasons (and I've found ways that `zstd` errors on Close), any `.Close()` fail, then you're out of luck, and end up with borked archives.

Refs: 
```
tarbrotli.go:82:	tbr.Tar.cleanupWrapFn = func() {
tarsz.go:83:	tsz.Tar.cleanupWrapFn = func() {
tarlz4.go:90:	tlz4.Tar.cleanupWrapFn = func() {
tarbz2.go:88:	tbz2.Tar.cleanupWrapFn = func() {
tarbz2.go:100:	tbz2.Tar.cleanupWrapFn = func() {
targz.go:95:	tgz.Tar.cleanupWrapFn = func() {
targz.go:111:	tgz.Tar.cleanupWrapFn = func() {
tarxz.go:85:	txz.Tar.cleanupWrapFn = func() {
tarzst.go:82:	tzst.Tar.cleanupWrapFn = func() {
tarzst.go:94:	tzst.Tar.cleanupWrapFn = func() {
```

## Proposed fix

Its use in:
```
tar.go:185:		defer t.cleanupWrapFn()
```

is not wired up to return an error from `t.cleanupWrapFn()`.

I propose something like:
```
func... (...) (err error) {
   ...
   defer func() {
      err2 := t.cleanupWrapFn()
      if err == nil {
          err = err2
      }
   }()
```

abourget:
The updated zstd library brings perf optimization, memory consumption optimizations, as well as better detection of certain corrupt archives (which we've hit).

Each commit is isolated and could be cherry-picked separately.  Here's a summary of each commit:

    Boost zstd library to version `1.8.6` which brings optimizations in terms
    of speed and memory use.

    Added support to set encoder and decoder options for zstd, necessary
    for some archive sizes (regarding memory usage).

    Fixes #184: handle errors coming out of (most) `Close()` statements that
    return errors.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -14,6 +14,8 @@ github.com/klauspost/compress v1.7.1 h1:VRD0WLa8rweLB7alA5WMSVkoAtrI8xou5RrNd4JU
 github.com/klauspost/compress v1.7.1/go.mod h1:RyIbtBH6LamlWaDj8nUwkbUhJ87Yi3uG0guNDohfE1A=
 github.com/klauspost/compress v1.7.4 h1:4UqAIzZ1Ns2epCTyJ1d2xMWvxtX+FNSCYWeOFogK9nc=
 github.com/klauspost/compress v1.7.4/go.mod h1:RyIbtBH6LamlWaDj8nUwkbUhJ87Yi3uG0guNDohfE1A=
+github.com/klauspost/compress v1.8.6 h1:970MQcQdxX7hfgc/aqmB4a3grW0ivUVV6i1TLkP8CiE=
+github.com/klauspost/compress v1.8.6/go.mod h1:RyIbtBH6LamlWaDj8nUwkbUhJ87Yi3uG0guNDohfE1A=
 github.com/klauspost/cpuid v1.2.0 h1:NMpwD2G9JSFOE1/TJjGSo5zG7Yb2bTe7eq1jH+irmeE=
 github.com/klauspost/cpuid v1.2.0/go.mod h1:Pj4uuM528wm8OyEC2QMXAi2YiTZ96dNQPGgoMS4s3ek=
 github.com/klauspost/pgzip v1.2.1 h1:oIPZROsWuPHpOdMVWLuJZXwgjhrW8r1yEX8UqMyeNHM=
--- a/tarzst.go
+++ b/tarzst.go
@@ -92,8 +92,8 @@ func (tzst *TarZstd) wrapWriter() {
 		zstdw, err = zstd.NewWriter(w, tzst.eopts...)
 		return zstdw, err
 	}
-	tzst.Tar.cleanupWrapFn = func() {
-		zstdw.Close()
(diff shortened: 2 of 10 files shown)


go.sum
```diff
@@ -14,6 +14,8 @@ github.com/klauspost/compress v1.7.1 h1:VRD0WLa8rweLB7alA5WMSVkoAtrI8xou5RrNd4JU
 github.com/klauspost/compress v1.7.1/go.mod h1:RyIbtBH6LamlWaDj8nUwkbUhJ87Yi3uG0guNDohfE1A=
 github.com/klauspost/compress v1.7.4 h1:4UqAIzZ1Ns2epCTyJ1d2xMWvxtX+FNSCYWeOFogK9nc=
 github.com/klauspost/compress v1.7.4/go.mod h1:RyIbtBH6LamlWaDj8nUwkbUhJ87Yi3uG0guNDohfE1A=
+github.com/klauspost/compress v1.8.6 h1:970MQcQdxX7hfgc/aqmB4a3grW0ivUVV6i1TLkP8CiE=
+github.com/klauspost/compress v1.8.6/go.mod h1:RyIbtBH6LamlWaDj8nUwkbUhJ87Yi3uG0guNDohfE1A=
```
mholt: Can you run `go mod tidy` so we no longer have old versions of this lib in our go.sum?

tarzst.go
```diff
@@ -22,6 +25,16 @@ func (*TarZstd) CheckExt(filename string) error {
 	return nil
 }
 
+// SetEncoderOptions allows you to configure the underlying zstandard encoder
+func (t *TarZstd) SetEncoderOptions(opts ...zstd.EOption) {
+	t.eopts = opts
+}
```
mholt: Why not just export the fields in the first place like all the existing config fields?

abourget: Right! Hadn't seen the other ones, even better.  What about `EncoderOptions []zstd.EOption` and `DecoderOptions` ?

abourget:
btw, the zstd library has a bug and creates corrupt archives (present since 1.7.4 apparently).. so I'm waiting for https://github.com/klauspost/compress/issues/170 to be solved before moving ahead here..

klauspost:
@abourget v1.9.0 released.

klauspost:
... And `v1.9.1` with a bugfix is also released.

mholt:
pinging @abourget

tarzst.go
mholt: Sounds good

PR #186, opened 2019-10-23
tarbz2 was added

kochurovro:
ByHeaders is in progress. tarbz2 was added


kochurovro:
Nice. Yes, I guess u can but I will continue working on it. Is it ok?

mholt:
I will wait until you are done then.

mholt:
Sorry I never got around to getting back to this... should be superseded by #302 though, wherein I completely rewrote the archiver package in preparation for v4. It has tar.bz2 support and also matches by headers. Thanks for contributing this though!

PR #190, opened 2019-11-14
Fix module name

### LINKED ISSUE #187: go get results in 410 Gone
#162  What version of the package or command are you using?
v3.1.1

## What are you trying to do?
I'm trying to upgrade to v3.2.0 using `go get -u github.com/mholt/archiver`

## What steps did you take?
First I ran `go get -u github.com/mholt/archiver` but it didn't find the newest version. The I tried to manually change the version to v3.2.0 in the `go.mod` file, but then it fails with an error instead.

I expected the new version to be found and downloaded, but the `go get` command fails with `verifying github.com/mholt/archiver@v3.2.0+incompatible/go.mod: github.com/mholt/archiver@v3.2.0+incompatible/go.mod: reading https://sum.golang.org/lookup/github.com/mholt/archiver@v3.2.0+incompatible: 410 Gone` 

From what I can read in issues in the the go repo (https://github.com/golang/go) they have some some problems with the `+incompatible` syntax in 1.13. Maybe you should look into switching to the `github.com/mholt/archiver/v3` import name convention to avoid the `+incompatible` all together?

nmiyake:
Update module name to have "/v3" suffix and update
imports and README accordingly.

Fixes #187

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ Tar files can optionally be compressed using any of the above compression format
 ## Install
 
 ```bash
-go get -u github.com/mholt/archiver/cmd/arc
+go get -u github.com/mholt/archiver/v3/cmd/arc
 ```
 
 Or download binaries from the [releases](https://github.com/mholt/archiver/releases) page.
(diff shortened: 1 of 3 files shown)


nmiyake:
Great, thanks!

Can you tag v3.2.1 or v3.3.0? Once that's done, this should be properly importable as a module!

nmiyake:
Filed #191 to track tagging the next release.

PR #193, opened 2019-11-14
go.mod: update github.com/andybalholm/brotli to v1.0.0

nmiyake:


nmiyake:
Last PR in this series! The maintainer for this repository agreed to tag the latest version (https://github.com/andybalholm/brotli/issues/9), so it's possible to use a versioned module dependency now. No code change for dependency.

This is the last non-tagged module dependency besides github.com/xi2/xz (and based on https://github.com/xi2/xz/pull/7, it looks like that repo is not actively maintained and unlikely to add modules/tags), so should be the last PR of this type for now :)

PR #197, opened 2019-11-25
Fix import path in test

nmiyake:


nmiyake:
This should fix the CI failures introduced by https://github.com/mholt/archiver/commit/14b27371bf503d5f94e087aa55c5d36799b6dc88.

PR #198, opened 2019-11-25
Add CI check to ensure module definition is correct

nmiyake:
Fail verification if "go mod tidy" results in any modifications
to go.mod or go.sum.


nmiyake:
I'm not very familiar with the Azure pipeline CI system, but semantically the diff I added here is likely the check that should be performed -- basically, we should ensure that "go mod tidy" has been run before code has been pushed. This would have caught the issue introduced by https://github.com/mholt/archiver/commit/14b27371bf503d5f94e087aa55c5d36799b6dc88.

The CI failures for this PR should be resolved once https://github.com/mholt/archiver/pull/197 is merged and this PR is updated to include it.

This manually diff after running "go mod tidy" is the best that can be done until https://github.com/golang/go/issues/27005 is implemented/available.

mholt:
Should we test this by merging master into this branch now that #197 is merged?

nmiyake:
Interesting, it looks like the redirect diff command (`diff -u <(echo -n) <(git diff go.mod)`) doesn't work properly for the Windows image. I'm not very familiar with that environment... Options are either:

1. Figure out way to perform this operation on Windows
2. Update CI configuration to skip/ignore this operation on Windows (the `go mod tidy` operation works over all build tags, so the execution environment doesn't matter and the Mac/Linux coverage should probably be sufficient)

(1) is the more "correct" approach, but will take longer since I'll have to iterate against this Azure environment and will mostly just be Googling and trying different things until I can get it to work :)

nmiyake:
OK, got this to work on the Windows image as well with only modifications to the CI file -- should be ready.

mholt:
Awesome, thanks for the effort! :+1:

PR #199, opened 2019-11-25
Add ReaderUnarchiver interface and provide implementations

### LINKED ISSUE #150: Method to extract zip archive from memory

nmiyake:
This commit adds a new ReaderUnarchive interface, which supports
performing an Unarchive operation from an io.Reader with a provided
size.

Fixes #150
Addresses #131

### DIFF
--- a/archiver.go
+++ b/archiver.go
@@ -78,6 +78,12 @@ type Unarchiver interface {
 	Unarchive(source, destination string) error
 }
 
+// ReaderUnarchiver is a type that can extract archive files
+// from an io.Reader into a folder.
+type ReaderUnarchiver interface {
+	ReaderUnarchive(source io.Reader, size int64, destination string) error
+}
+
 // Writer can write discrete byte streams of files to
 // an output stream.
 type Writer interface {
--- a/archiver_test.go
+++ b/archiver_test.go
@@ -331,6 +331,22 @@ func testArchiveUnarchive(t *testing.T, au archiverUnarchiver) {
 
 	// Check that what was extracted is what was compressed
 	symmetricTest(t, auStr, dest, true, true)
+
+	// Read outfile bytes and create a reader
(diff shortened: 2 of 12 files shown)


nmiyake:
Will call out the following:

* Not sure if `ReaderUnarchiver` and `ReaderUnarchive` are the best interface/function names here -- tried to match the existing style
* If `ImplicitTopLevelFolder` is set to true, these implementations will fully read the reader bytes into memory first, which is not very memory-efficient. However, this is explicitly called out in the documentation, and not sure that this workflow is very common
* If `ImplicitTopLevelFolder` is set to true, the implicit top-level folder is hard-coded to be named "archive" -- the file logic bases the name on the input filename, but there's no such analog when using an input reader

This is somewhat related to #131 -- as a follow-on, I'd like to add a parallel interface/functions for writing an archive to an `io.Writer`, so the eventual naming there should likely be consistent (for example, `WriterArchiver` and `WriterArchive`?).

tabboud:
@mholt Is this something you would consider merging?

mholt:
Thank you for this! Sorry I never got back to this, it's been a crazy year.

This PR will be superseded by #302, which is a complete rewrite of archiver that has stream-oriented APIs. I think you'll like it. Let me know if you have any feedback! Really appreciate your contribution.

PR #200, opened 2019-11-25
Add WriterArchiver interface and provide implementations

### LINKED ISSUE #131: Why did Compress and Uncompress disappear?

nmiyake:
This commit adds a new WriterUnarchiver interface, which supports
performing an Archive operation to an io.Writer.

Fixes #131 when paired with #199

### DIFF
--- a/archiver.go
+++ b/archiver.go
@@ -67,6 +67,13 @@ type Archiver interface {
 	Archive(sources []string, destination string) error
 }
 
+// WriterArchiver is a type that can create an archive
+// from a list of source file names and write it to an
+// io.Writer.
+type WriterArchiver interface {
+	WriterArchive(sources []string, destination io.Writer) error
+}
+
 // ExtensionChecker validates file extensions
 type ExtensionChecker interface {
 	CheckExt(name string) error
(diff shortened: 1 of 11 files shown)


nmiyake:
This PR is the analog to #199 for creating archives. Happy to combine them together or keep them separate.

Also, if you want to provide high-level feedback before delving into PRs totally open to that as well.

mholt:
Thank you! As with #199, closing as this is superseded by #302, which features new core APIs that are stream-oriented.

PR #201, opened 2019-11-29
Fix issue when reading symlink during tar Archive

### LINKED ISSUE #202: Creating tar archive fails with symbolic links
## What version of the package or command are you using?
3.3.0

## What are you trying to do?
Create a new tar archive.

## What steps did you take?
Create a tar archive from the following:
```
$ ls -l /tmp/testdata/
total 4
-rw-rw-r-- 1 1 Dec  2 09:22 file1
lrwxrwxrwx 1 9 Dec  2 09:22 linktofile1 -> /tmp/testdata/file1
```
With:
```go
package main

import (
        "log"

        "github.com/mholt/archiver/v3"
)

func main() {
        t := archiver.NewTar()
       err := t.Archive(
                []string{
                        "/tmp/testdata/file1",
                        "/tmp/testdata/linktofile1",
                },
                "/tmp/archive.tar")
        if err != nil {
                log.Fatalf("failed to create archive: %v", err)
        }
}
```

A new archive is created at `/tmp/archiver.tar` but you get an error:
```
2019/12/02 09:23:35 failed to create archive: walking /tmp/testdata/linktofile1: /tmp/testdata/linktofile1: writing: linktofile1: readlink: readlink linktofile1: no such file or directory
```

When reading symbolic links using `os.Readlink`, a full filesystem path should be given rather than just the link name.

## Please link to any related issues, pull requests, and/or discussion
#201

technicianted:
This PR fixes an issue when create tar archives with symbolic links. 

The code uses the `Name()` value to call `os.Readlink()`. This does not work since this value is only the name of the link. not the full path.

To resolve this, we need to pass down the original full FS path and use it during `os.*` calls.

Fixes #202

### DIFF
--- a/archiver.go
+++ b/archiver.go
@@ -105,6 +105,9 @@ type Extractor interface {
 type File struct {
 	os.FileInfo
 
+	// OriginalPath is the original full path to file
+	OriginalPath string
+
 	// The original header info; depends on
 	// type of archive -- could be nil, too.
 	Header interface{}
--- a/tar.go
+++ b/tar.go
@@ -295,6 +295,7 @@ func (t *Tar) writeWalk(source, topLevelFolder, destination string) error {
 				FileInfo:   info,
 				CustomName: nameInArchive,
 			},
+			OriginalPath: fpath,
 			ReadCloser: file,
 		})
 		if err != nil {
@@ -340,7 +341,7 @@ func (t *Tar) Write(f File) error {
 	var linkTarget string
 	if isSymlink(f) {
 		var err error
-		linkTarget, err = os.Readlink(f.Name())
+		linkTarget, err = os.Readlink(f.OriginalPath)
 		if err != nil {
 			return fmt.Errorf("%s: readlink: %v", f.Name(), err)
 		}


mholt:
Thanks. Is there a test case that exhibits the bug? Or a related issue already reported?

technicianted:
#202

mholt:
Ah, thanks. I'm willing to merge this, but first, would it be an easy task to add a file to the test corpus which exhibits the behavior? If it is too complicated or something, let me know, but it'd be nice to be able to verify that the fix is working with a test.

nishant-handa:
can we get this fix merged please.

kneal:
@mholt Any updates on this one?

mholt:
Nope, sorry, if there were updates they would be posted here.

I've been busy readying the Caddy 2 release. In the meantime, this PR needs tests, as stated above.

jbrockopp:
@technicianted @mholt 

I've opened a separate PR that attempts to copy the changes made in this PR, but also add tests:

https://github.com/mholt/archiver/pull/230

Please let me know if anything else is required 👍

mholt:
Probably fixed in #296.

PR #205, opened 2019-12-12
Replace Zip and Flate packages.

calavera:
Use the optimized versions from github.com/klauspost/compress.
In smoke tests, compressing 7K files took 3 seconds, rather than 12
seconds with the official packages.

### DIFF
--- a/zip.go
+++ b/zip.go
@@ -1,9 +1,7 @@
 package archiver
 
 import (
-	"archive/zip"
 	"bytes"
-	"compress/flate"
 	"fmt"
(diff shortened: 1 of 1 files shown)


coolaj86:
@calavera Is this still relevant? Are the official packages still slow?

Also, what's the use case where this is a problem?

I feel like `arc` feels a niche of convenience, and we really value being light on dependencies, so I'm not sure if we want to bring those in or not. I'm open to it, but cautious.

Can we open this back up for discussion?

PR #208, opened 2020-01-25
#118: ByHeader support for gzip, bzip2, and xz

zaquestion:
Not complete support, but a start.

Open questions:
- Is it reasonable to change the function signature of `ByHeader` to match `ByExtension`?
- To identify archive+compression type (tar.gz) should we go so far as to decompress files match in by a compression to see if the underlying file type is an archive?
- h2non/filetype checks are pretty simply (only checking the magic numbers), we could potentially implement inline instead of bringing in the dependency


mholt:
Thank you for this. Sorry the PR went unanswered for so long. I was busy working on Caddy 2 at the time, and it was very busy and stressful.

This PR will be superseded by #302, which has "ByHeader" support for all file formats (but the mechanism is called something else, "Match", and is implemented differently and natively in-package). Thanks again!

Issue #130, opened 2018-12-12
Return the archived/unarchived file path


yangyadi1993:
## What would you like to have changed?

Currently, only the error message will be returned when we unarchived files. I don't know what the archived file path is. If the archived path could be returned, it will be more user friendly.

Just asking if you have the plans :)

mholt:
I don't fully understand what you mean. It's not specific enough. If you could fill out the full issue template, complete with code examples in this case that would be helpful, thanks.

J7mbo:
@mholt  I think that this issue is related to [this](https://github.com/golang/go/issues/32300).

I'm looping around a zip of a bitbucket repo. I'm trying to extract all text files that are mine and **not** in vendor. 

When looking at your `FileInfo` provided from `Walk()`, getting to a vendor/ directory, the `Name()` function only contains the name of the file, not the path itself (which is weird because when debugging I can see the path). Also `FileInfo` when printed only contains the name of the file too.

As a result, I can't exclude the text files from vendor with a `strings.Contains()`

Any chance that the path of the file can be included alongside the `FileInfo`? I'm happy to create a new issue but would rather know if there's the option for this first.

ZaynJarvis:
like the filepath.Walk
https://golang.org/pkg/path/filepath/#WalkFunc

mholt:
I think this is implemented in #302, as NameInArchive, if I understand correctly.

PR #215, opened 2020-02-23
Added FullPath to File for Zips.

J7mbo:
Hi @mholt, saw that you're busy with Caddy, all good I'll use my fork for a while. 

Related to #130.

When you have time, can this be considered? There's a lot missing, the `File` is used everywhere and I only needed it when walking around a Zip, but perhaps we can discuss making this work for everything and get the PR in.

### DIFF
--- a/archiver.go
+++ b/archiver.go
@@ -103,6 +103,8 @@ type Extractor interface {
 // File provides methods for accessing information about
 // or contents of a file within an archive.
 type File struct {
+	FullPath string
+
 	os.FileInfo
 
 	// The original header info; depends on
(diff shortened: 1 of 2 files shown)


J7mbo:
If this isn't going anywhere I will close.

archiver.go
```diff
@@ -103,6 +103,8 @@ type Extractor interface {
 // File provides methods for accessing information about
 // or contents of a file within an archive.
 type File struct {
+	FullPath string
```
mholt, line 106: Would be useful to provide a comment here explaining what the field should contain.

mholt (changes requested):
Thanks, sorry for the delay.

mholt:
I think this is implemented in #302, as NameInArchive, if I understand correctly.

J7mbo:
All good, I didn't put too much effort into it. Good luck in 2022! 👋

PR #223, opened 2020-06-14
Support decompression compression algorithm using zstd/lzma/bzip2 zip file.

fcharlie:
According to the latest zip file format specification [APPNOTE.TXT](https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT), zstd has been accepted as the algorithm supported by zip, the number is 20, this PR gives people to create and decompress zip files using zstd compression algorithm.

Follow-up of other compression software: 
https://github.com/mcmilk/7-Zip-zstd/issues/132
https://github.com/szcnick/p7zip/issues/25

Restriction: When creating a zip(zstd) file, setting the compression level is not supported


zip.go
```diff
@@ -6,11 +6,26 @@ import (
...
+)
+
+// Compression methods.
+const (
+	Store      uint16 = 0  // no compression
+	Deflate    uint16 = 8  // DEFLATE compressed
+	BZIP2      uint16 = 12 // bzip2
+	LZMA       uint16 = 14 //LZMA
+	ZSTD       uint16 = 20 //see https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT.
+	WINZIPZSTD uint16 = 93
```
mholt, line 28: This package is shared by many types, so these constant names should probably be prefixed by something like `Zip` since they're specific to the zip archiver.

They should also be a custom type perhaps, not just "any" `uint16`, e.g. `type ZipCompressionMethod uint16`.

Also clean up the comments (spacing).

zip.go
```diff
@@ -55,9 +70,12 @@ type Zip struct {
 	// the operation will continue on remaining files.
 	ContinueOnError bool
 
-	zw   *zip.Writer
-	zr   *zip.Reader
-	ridx int
+	//Compression algorithm
```
mholt, line 73:
```suggestion
	// Compression algorithm
```

zip.go
```diff
@@ -145,7 +201,6 @@ func (z *Zip) Unarchive(source, destination string) error {
 		return fmt.Errorf("opening zip archive for reading: %v", err)
 	}
 	defer z.Close()
-
```
mholt, line 148: ?

PR #226, opened 2020-07-06
Disable ZIP+LZMA

fcharlie:
7z seems to have adopted a different strategy to create ZIP+lzma. 

Here also fixes archive ls panic, unsupported compression method panic

### DIFF
--- a/zip.go
+++ b/zip.go
@@ -108,13 +108,6 @@ func registerDecompressor(zr *zip.Reader) {
 		}
 		return bz2r
 	})
-	// zr.RegisterDecompressor(uint16(LZMA), func(r io.Reader) io.ReadCloser {
-	// 	lr, err := lzma.NewReader(r)
-	// 	if err != nil {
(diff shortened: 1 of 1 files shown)


zip.go
```diff
@@ -348,14 +353,18 @@ func (z *Zip) Create(out io.Writer) error {
...
 			return bzip2.NewWriter(out, &bzip2.WriterConfig{Level: z.CompressionLevel})
 		})
-	case LZMA:
-		z.zw.RegisterCompressor(uint16(LZMA), func(out io.Writer) (io.WriteCloser, error) {
-			return lzma.NewWriter(out)
-		})
+	// case LZMA:
+	// 	z.zw.RegisterCompressor(uint16(LZMA), func(out io.Writer) (io.WriteCloser, error) {
+	// 		return lzma.NewWriter(out)
+	// 	})
```
mholt, line 359: This can be removed entirely

PR #233, opened 2020-09-19
use 'go get' because 'go install' does not get

coolaj86:
Oddly enough `go install` will not `go get`, but `go get` will `go install`, and so 'go get' does what is intended (download and install) whereas 'go install' does not (it only builds and installs from cache).


coolaj86:
Fixed in https://github.com/mholt/archiver/commit/45ba413222e433902bfac0f63eae69c285357cc3

PR #234, opened 2020-09-23
Cleanup for CI/CD Pipeline

coolaj86:
- update `.gitignore` for `arc` and vim
- add `tools` package for tooling dependencies (i.e. the linter)
- lock all dependencies in `go.mod`
- fix linter errors
- update azure pipeline
    - Update macOS to Mojave
    - Update Ubuntu to 18.04
    - Use BSD curl and tar to speed up Windows builds


mholt:
Oh, I'm not sure how I feel about adding almost a million lines of code to this repo... (in fact we tried vendoring in Caddy and decided it was a mistake). Go modules can pin the versions for us without all that extra baggage. Otherwise, rest of the bullet points look good.

coolaj86:
_Drastically_ speeds up CI/CD times.

I'd like to hear about why you decided that was a mistake with caddy.

That said, once I've got the CI/CD working again I can reset and commit without the `-mod=vendor` stuff.

mholt:
It's a long story. And our Caddy CI tests run in under a couple minutes usually without vendoring; I think with a cache of some sort. And that's a larger code base with way more tests. So, let's hold off on vendoring for now (easy to do later, harder to undo); the rest of the cleanup is good :)

coolaj86:
... and the Windows CI time is down to 3m. What can I say, I'm a wizard. 🧙‍♂️ 😄

cmd/arc/main.go
```diff
@@ -219,7 +219,7 @@ func getFormat(subcommand string) (interface{}, error) {
 		v.ContinueOnError = continueOnError
 		v.Password = os.Getenv("ARCHIVE_PASSWORD")
 	case *archiver.Tar:
-		v = mytar
+		f = mytar
```
coolaj86, line 222: @mholt I'd like your comment on this, please.

Before it was a no-op. Now it's acting as intended, and the tests are passing, but I'm not sure of any potential consequences. I'd be happy to revert this if there's any concern.

coolaj86:
@mholt All changes are complete and tested. This is ready for your review.

cmd/arc/main.go
mholt: This is fine, thanks! I think it's effectively equivalent since the value is a pointer.

zip.go
```diff
@@ -76,11 +76,11 @@ type Zip struct {
...
-	FileMethod     ZipCompressionMethod
-	zw             *zip.Writer
-	zr             *zip.Reader
-	ridx           int
-	decinitialized bool
+	FileMethod ZipCompressionMethod
+	zw         *zip.Writer
+	zr         *zip.Reader
+	ridx       int
+	//decinitialized bool
```
mholt, line 83: Leaving this here?

mholt (approved):
Awesome, this LGTM! Thanks so much!

Issue #196, opened 2019-11-16
Version subcommand 


Croissong:
## What would you like to have changed?
Provide an `archiver version` subcommand to display its current version corresponding to the github release tag.

## Why is this feature a useful, necessary, and/or important addition to this project?
I don't believe it's currently possible to figure out the used version, or is there? 

Thanks in advance :)

coolaj86:
@mholt I'd be willing to add this such that `arc version` outputs a version number.

Given no other direction I would use either `goreleases` or copy https://git.rootprojects.org/root/go-gitver/src/branch/master/gitver/gitver.go into `tools/gitver.go` and add a `go generate` command.

However, I know that you are particular about how things are done, so I'd like to solicit your take on the matter as to how this could be implemented to your liking.

coolaj86:
I suggest this subcommand:

```bash
arc version
```

And this as the output:

```txt
arc (archiver) v3.3.0 (g3719b73) 2020-07-17T21:03:06 2020
```

Dean-Coakley:
@coolaj86 This can be closed too?

Implemented in: https://github.com/mholt/archiver/pull/236

Croissong:
I assume it can, since @coolaj86 implemented it himself. Thanks a bunch you guys :)

PR #236, opened 2020-09-24
Updates for Releases ( and Goreleaser )

coolaj86:
- add `version` subcommand
- add `goreleaser` with bare binaries
- cleanup `README.md` with `prettier`
- add `webi` as install method
- clarify manual build and install instructions

Updated README can be previewed at https://github.com/mholt/archiver/tree/goreleaser

`goreleaser` is typically used with `tar` and `zip` but, being that we might assume the person that wants to install `arc` may not have that support on their system, it may make sense to not compress and package it.

~This may rely on https://github.com/goreleaser/goreleaser/pull/1821 and  https://github.com/goreleaser/goreleaser/pull/1823~

That said, Linux and Unix (Darwin) systems that I've encountered have GUI and CLI `tar.xz` support built-in, and Windows 10 has zip support both in the GUI and CLI (both in BSD tar.exe and Powershell), so it may be fair to package it as well.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,4 @@
+{
+  "bracketSpacing": true,
+  "printWidth": 120,
+}
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,0 +1,41 @@
+# This is an example goreleaser.yaml file with some sane defaults.
+# Make sure to check the documentation at http://goreleaser.com
+project_name: arc
+before:
+  hooks:
+    # You may remove this if you don't use go modules.
+    - go mod download
+    # you may remove this if you don't need go generate
+    - go generate ./...
+builds:
+  -
+    env:
+      - CGO_ENABLED=0
+    main: ./cmd/arc
+    goos:
+      - linux
+      - windows
+      - darwin
+    goarch:
+      - 386
+      - amd64
+      - arm
+      - arm64
+    goarm:
+      - 6
+      - 7
+archives:
+  -
+    format: binary
+    replacements:
+      darwin: mac
+checksum:
+  name_template: 'checksums.txt'
+snapshot:
+  name_template: "{{ .Tag }}-next"
+changelog:
+  sort: asc
+  filters:
+    exclude:
+      - '^docs:'
+      - '^test:'
--- a/README.md
+++ b/README.md
@@ -48,20 +48,75 @@ Files are put into the root of the archive; directories are recursively added, p
 
 Tar files can optionally be compressed using any of the above compression formats.
 
+## GoDoc
+
+See <https://pkg.go.dev/github.com/mholt/archiver/v3>
+
 ## Install
+
+### With webi
+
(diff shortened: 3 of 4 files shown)


coolaj86:
@mholt Using `goreleaser` requires a `GITHUB_TOKEN` to be able to publish to Github Releases.

Can you give me Github Release Publish access?

mholt:
@coolaj86 Sure, is that just a matter of creating a secret named GITHUB_TOKEN on this page? https://github.com/mholt/archiver/settings/secrets

coolaj86:
@mholt Reverse.

I believe I need to be bumped to "Maintainer" at https://github.com/mholt/archiver/settings/access and then I can create the token I need to use.

Using one of your secrets on that page could work, but only for a goreleaser CI/CD pipeline.

Since I want to add 7z support I need a cross-compiler toolchain that I have set up locally, but I don't know how I'd set up in Github Actions.

mholt:
@coolaj86 There is no option, it just says "Collaborator" and the only thing I can do is delete you. Want me to create the token for you? Just let me know what to do.

coolaj86:
@mholt Hmmm... because this is a personal account, not an organization, you don't get maintainers.

The problem is that personal tokens cannot be scoped to a repository. If you give me one of your tokens, I'll have access to all of your repositories, and I don't really want that.

I'll let it ruminate and see if I can come up with a clever solution. The first thing that comes to mind is publishing the releases to a separate, organizational repository.

mholt:
I see... keep me posted. CI might not be a bad option then?

coolaj86:
Actually, according to this ( https://docs.github.com/en/free-pro-team@latest/github/setting-up-and-managing-organizations-and-teams/repository-permission-levels-for-an-organization ) I should be able to create a release.

Huh. Yep. I can.

I just thought I couldn't because I couldn't see settings and I must have had my window previously at a size that it didn't show the `edit` button or something on the releases page.

coolaj86:
@mholt This is now ready for review + approval.

The goreleaser issue I cross-referenced is not a blocker.

.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
+  "bracketSpacing": true,
+  "printWidth": 120,
+}
```
mholt, line 4: Did you mean to commit this? Looks like an accident. (This appears to be a user-specific preference file. Probably shouldn't be committed.)

coolaj86: This is for the README.md so that any future changes to it follow the same style. I don't care what the style is, I just care that it's documented.

coolaj86:
```bash
prettier --write README.md
```

mholt: What tools support this file though? It's the first I've heard of it...

coolaj86: https://prettier.io

You don't worry about this much in the Go world because you've got `go fmt`, but `prettier` is an all-in-one for docs, js, python, html, css, etc, etc.

mholt (approved):
LGTM, thanks for helping clean things up!

a8nJ:
https://github.com/mholt/archiver/issues/196

PR #239, opened 2020-09-28
Add a way to strip top-level folder

### LINKED ISSUE #238: strip-components support
Add a way to strip top-level directories like GNU and BSD tar does

In some cases this feature is needed to extract actual files without top-level directories. Would be awesome to have this feature natively :+1:

None (yet)

## Please link to any relevant issues, pull requests, or other discussions.
POSIX shell implementation: https://github.com/kisslinux/kiss/blob/fc54435595bed7db083a358bd673f520b70447aa/kiss#L301

illiliti:
Closes #238

### DIFF
--- a/archiver_test.go
+++ b/archiver_test.go
@@ -452,6 +452,60 @@ func symmetricTest(t *testing.T, formatName, dest string, testSymlinks, testMode
 	}
 }
 
+func TestUnarchiveWithStripComponents(t *testing.T) {
+	testArchives := []string{
+		"testdata/sample.rar",
+		"testdata/testarchives/evilarchives/evil.zip",
+		"testdata/testarchives/evilarchives/evil.tar",
+		"testdata/testarchives/evilarchives/evil.tar.gz",
+		"testdata/testarchives/evilarchives/evil.tar.bz2",
+	}
+
+	to := "testdata/testarchives/destarchives/"
+
+	for _, archiveName := range testArchives {
+		f, err := ByExtension(archiveName)
+
+		if err != nil {
+			t.Error(err)
+		}
+
+		var target string
+
+		switch v := f.(type) {
+		case *Rar:
+			v.OverwriteExisting = false
+			v.ImplicitTopLevelFolder = false
+			v.StripComponents = 1
+			target = "quote1.txt"
+		case *Zip:
+		case *Tar:
+			v.OverwriteExisting = false
+			v.ImplicitTopLevelFolder = false
+			v.StripComponents = 1
+			target = "safefile"
+		case *TarGz:
+		case *TarBz2:
+			v.Tar.OverwriteExisting = false
+			v.Tar.ImplicitTopLevelFolder = false
+			v.Tar.StripComponents = 1
+			target = "safefile"
+		}
+
+		u := f.(Unarchiver)
+
+		if err := u.Unarchive(archiveName, to); err != nil {
+			fmt.Println(err)
+		}
+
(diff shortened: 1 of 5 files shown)


coolaj86:
Thanks @illiliti, I like this. 

Would you minding changing it to work just like `--strip-components`? We could call it `StripComponents` or `StripParentDirs` or `StripDepth` or something like that, as an int, and then update the code to strip a variable number of directories accordingly?

We'd also need to update the command with an appropriate flag.

illiliti:
@coolaj86 I did it, hope in the right way

coolaj86:
Tar looks great to me. Can you also implement `Zip` and `Rar` so that it's consistent? (having the option only work for `tar` would be confusing)

There are three places that will need this change:

- [x] Tar
- [ ] Zip
- [ ] Rar

## Testing

- [ ] Manual Test / Sanity Check
- [ ] Automated Test

### Manual / Sanity Check

Have you done any testing with it from the command line? Such as with the `testdata` files?

### Automated / Go

I think an automated test for this should be fairly simple to produce.

The test would be very similar to `TestSafeExtraction` but, instead of using evil files, we'd need the stripComponents option turned on.

See https://github.com/mholt/archiver/blob/master/archiver_test.go#L456

It would need an Unarchive function similar to this https://github.com/mholt/archiver/blob/master/archiver.go#L206, but perhaps with a type `switch` statement so that the `StripComponents` could be set on each.

illiliti:
> Can you also implement `Zip` and `Rar` so that it's consistent?

Done

By the way, while implementing strip-components to zip and rar, i realized that API is very inconsistent. Just compare `untarNext`, `unrarNext` and `extractNext` functions and you will understand what i mean.


> Have you done any testing with it from the command line? Such as with the `testdata` files?

Yep. Current functionality is fully identical to GNU tar.

> I think an automated test for this should be fairly simple to produce.

Will do

illiliti:
@coolaj86 Everything is done

coolaj86:
>  i realized that API is very inconsistent

Yes, this is the reason for `Arc`s existence - to normalize away some of those differences. By virtue of the fact that the people who created `zip`, `tar`, and `rar` had very different and incongruent opinions about "the best way" to archive files, we have nasty bits to work around.

coolaj86:
This all looks great. I'll do a little manual testing myself and merge this in.

There are two other patch PRs likely to come in in the next few days, so I'll publish the v3.4 release once I have those as well.

coolaj86:
Just popping in to let you know I haven't forgotten. I'm waiting on the rebase of one more soft-approved PR and then I'll merge this and cut a new release.

illiliti:
Okay, thank you !

coolaj86 (approved):
LGTM

89z:
@illiliti thanks for writing this pull request.

I have a question, I am looking at this code you wrote:

https://github.com/mholt/archiver/blob/074da64920d3c1188f608a92eceb61e83b31317b/tar.go#L246-L249

instead of a loop, I was wondering why you didnt do something like this:

~~~go
package main
import "strings"

func strip(s string, n int) string {
   a := strings.SplitN(s, "/", n + 1)
   return a[n]
}

func main() {
   s := "year/month/day/hour"
   s1 := strip(s, 1)
   s2 := strip(s, 2)
   println(s1, s2)
}
~~~

probably doesnt make much difference either way, just curious. Thanks again!

illiliti:
I wasn't familiar with Go Std Library at that moment. That's why I decided to use C-style for loop. Feel free to submit PR if you think that your approach is more idiomatic and correct.

PR #246, opened 2020-10-03
Added IllegalPathError

ibraimgm:
Created a `IllegalPathError` struct, as well as a the `IsIllegalPathError` helper function.
Note that the struct has an additional field, `AbsolutePath`, that is **not** used for printing, but might be useful for for debugging purposes to library users.

### DIFF
--- a/error.go
+++ b/error.go
@@ -1,0 +1,27 @@
+package archiver
+
+import (
+	"fmt"
+	"strings"
+)
+
+// IllegalPathError is an error returned when an illegal
+// path is detected during the archival process.
+//
+// By default, only the Filename is showed on error, but you might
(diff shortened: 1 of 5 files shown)


coolaj86:
Thanks @ibraimgm. This looks good to me.
There is one suggestion I have, but it's not a blocker:

Rather than checking the string, I believe you should be able to directly check the type:

```go
_, ok := err.(ErrInvalidPath)
return ok
```

What do you think?

coolaj86:
I'm going to go ahead and merge this in as-is so that it's not blocking others. I'll open a new issue for using type checking over string checking.

coolaj86 (approved):
LGTM.

Noted that we want to prefer type checking over string checking.

Issue #247, opened 2020-10-04
craft zip file for symlink testing


coolaj86:
We need a special zip file that cannot be created with normal commandline tools. It requires crafting with an API. This should be possible with [`archive/zip#Writer`](https://golang.org/pkg/archive/zip/#Writer), for example.

We want a **double entry** of a file - the first being a symlink such that the second will be placed in an arbitrary location:

```txt
./goodfile.txt  "hello world"         (file)
./bad/file.txt  => ../../badfile.txt  (symlink)
./bad/file.txt  "Mwa-ha-ha"           (file)
./morefile.txt  "hello world"         (file)
```

This should go in [testdata/testarchives/evilarchives/](https://github.com/mholt/archiver/tree/master/testdata/testarchives/evilarchives) as `double-evil.zip` and `double-evil.tar` (if it is allowed).

See also https://github.com/mholt/archiver/issues/242#issuecomment-703086020

kross9924:
can I work on this?

coolaj86:
Assigned!

kross9924:
@coolaj86 can you please elaborate on zip file structure.

coolaj86:
@kross9924 Can you ask a more specific question? What about the file structure with 4 files as described above is unclear?

kross9924:
Is this file structure ok  ?

                                                                                      double-evil.zip
                                                                                       /                   \
                                                                            badfile.txt                dir
                                                                                                               \
                                                                                                                  subdir     
                                                                                                             /         \                 \
                                                                                                  goodfile.txt   morefile.txt   bad
                                                                                                                                              \
                                                                                                                                           file.txt

kross9924:
Is this structure correct>

double-evil.zip
1.badfile.txt  2.dir
                       2.1 subdir
                          2.1.1 goodfile.txt     2.1.2 morefile.txt     2.1.3 bad
                                                                                           2.1.3.1 file.txt

coolaj86:
Use code fences and tables:

~~~md
```bash
lsd --tree .
```

```txt
./
├── bad/
│  └── file.txt
├── goodfile.txt
└── morefile.txt
```

| filepath | type | contents |  link |
|  ------- | ---- | --------- | ---- |
| ./goodfile.txt | file | "hello world"  | N/A |
| ./bad/file.txt | symlink | N/A | ../../badfile.txt |
| ./bad/file.txt | file | "Mwa-ha-ha"  | N/A |
[...]

PR #253, opened 2020-10-13
add double-evil.tar and double-evil.zip inside testdata/testarchives/evilarchives

kross9924:
#247


kross9924:
When will it get accepted @coolaj86 ?

coolaj86:
@kross9924 Will you update the PR to include the code used to create these as well?

I think putting them in `testdata/create-evil-tar.go` and `testdata/create-evil-zip.go` might be a good choice.

Issue #183, opened 2019-10-17
Compress Zstd defaults to low memory


fproulx-dfuse:
The zstd library defaults to low memory mode which is very conservative with my use case. I had to fork your lib to disable this low memory mode (I use `WithDecoderLowmem(false)`)

Consider disabling this or set to more usable value...

mholt:
We'd be happy to consider a pull request that makes it configurable. Should be an easy enhancement.

PR #254, opened 2020-10-17
Add exported encoder/decoder opts for zstd

Dean-Coakley:
Fixes: https://github.com/mholt/archiver/issues/183

Allows setting of any encoder or decoder options when imported.

In the referenced issue `zstd.WithDecoderLowmem(false)` could be set for example.


coolaj86:
Makes perfect sense to me. Thanks.

coolaj86 (approved):

Issue #180, opened 2019-09-01
Way to avoid pulling every dependency


kevinburke1:
I am using archiver to pack and unpack .tar.gz files. For a project I am working on I am vendoring every library with `go get -mod=vendor ./...`. Unfortunately with archiver this means that I also must vendor the Brotli, Snappy etc. upstream libraries, even though the binary I am using doesn't actually link them or need them. This means an unnecessarily large vendor directory, larger Git pushes/pulls, etc.

Is there a way to only get the Targz struct out of this package? That way I wouldn't need to depend on/download the Brotli library.

mholt:
How large is your vendor folder? My recommendation would simply be to fork the project and remove all the files for the formats you don't need.

kevinburkemeter:
It's over 10MB at this point... one approach would be to mirror e.g. crypto/sha{1,256,512} where multiple different packages implement a single interface, and callers can either use the broad interface or the single package that they need.

mholt:
Okay, you're referring to a plugin model kind of like what [Caddy](https://github.com/caddyserver/caddy) does.

I don't have the time to work on that for this repository right now, but if it was a PR I could review it. (I am not sure how quickly I could get around to it, though.) The quickest short-term solution for you would probably be to do what I suggested and fork it, remove the features you don't need, and just use that for now.

kevinburkemeter:
I'm imagining something where each different format (brotli, zip etc) would have its own package, and then the top-level Archiver could either import them via aliases or call Register() or something to get all of them. Most people could import Archiver, if you cared you could import the package you want directly and avoid the other dependencies.

I took a short stab at this and two problems come up right away:

- everything depends on `*archiver.File` and `*archiver.FileInfo` so if the subpackages have to import those from the main package, you're toast since then you get everything anyway. We could also move those to a subpackage - say, `common` or something - to avoid a circular import.
- same with the helpers like `fileExists`, which would also need to move to a subpackage.

kevinburkemeter:
I took a stab at laying out what this could look like, let me know what you think.

https://github.com/mholt/archiver/compare/master...kevinburkemeter:package-layout?expand=1

mholt:
Ah, yeah, so you want something like what Caddy does. Totally doable for next major version, _but_ there is a high cost: it's a lot of work to build the application, since you essentially have to customize it for your own requirements. Then you have to answer questions like, what would the default/standard distribution look like: which packages are included, etc?

[...]

PR #258, opened 2020-10-31
all: split zip out into subpackage

kevinburke1:
Building on the discussion in #180, this is a way for callers to be
able to use the zip package directly without importing the other
compression/decompression packages. It uses aliases to preserve
backwards compatibility.

The utility functions and types used by the zip package needed to
be split out into a subpackage as well (named "common") to avoid a
circular dependency issue.


Jacalz:
I think this is a great change to make. Just want to give a friendly heads up that the tests seem to be failing on the PR.

kevinburkemeter:
Looking, thanks.

kevinburkemeter:
I can reproduce _an_ test failure locally but just a heads up that I don't have permission to view the Azure build logs.

kevinburkemeter:
Tests fixed! And also made the PR up to date.

Jacalz (changes requested):
Sorry, but this would need to be rebased on the work in https://github.com/mholt/archiver/commit/699ca560fed0e13b9c0639b01e372f19816ccc8c, to use github.com/klauspost/compress/zip for the zip handling. It is also out of date with the master branch.

Hopefully @mholt can review this once that is fixed :slightly_smiling_face:

kevinburkemeter:
Please check it now

mholt:
Thanks, I will review when I have a chance! Currently pretty focused on another project.

Jacalz:
I will make sure to test this as well when I get some time to do so. Will be interesting to see how the binary size is affected.

Jacalz (approved):
Nice work! This looks good to me. I am seeing around a 0.5MB decrease in binary size when using this over the standard `archiver` package (Go1.15 on Linux). Not massive, but a good step in the right direction :slightly_smiling_face:

Jacalz:
We should probably move all the different formats out to their own packages, but this is a great step on the way there :slightly_smiling_face:

mholt:
Hey, thanks for this PR. I remember I just commented again on the related issue today. It's not a change I feel compelled to make quite yet, but if more compelling use cases / requests come in for it, I might be willing to consider it.

Closing, as this PR is obsoleted by #302, which has true "plugin" support for the formats. So if we do decide to do it, it should be a little easier.

PR #263, opened 2020-11-15
cmd: Clean up flag parsing

89z:
- Since "flag" is already being used, use it as well for "help" and "version"
  flags
- When printing usage string, do not add extra newline
- Utilize "flag.NArg" to determine remaining arguments
- When running "fatal", only add newline as needed
- Rename "fatal" function for consistency
- Update help text to reflect flag changes

### DIFF
--- a/cmd/arc/main.go
+++ b/cmd/arc/main.go
@@ -24,6 +24,8 @@ var (
 	stripComponents        int
 	continueOnError        bool
 	specifyFileType        string
+	showHelp bool
+	showVersion bool
 )
 
 var (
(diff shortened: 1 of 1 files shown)


cmd/arc/main.go
```diff
@@ -24,6 +24,8 @@ var (
 	stripComponents        int
 	continueOnError        bool
 	specifyFileType        string
+	showHelp bool
+	showVersion bool
```
Jacalz, line 28: It looks like these two are not set up and formatted correctly.

Jacalz (changes requested):

cmd/arc/main.go
89z: I am not doing another commit just to add some spaces or whatever.

Jacalz: Well, I certainly wouldn’t approve a PR with incorrect formatting. Just run `go fmt` on the file and it will sort out the formatting for you. This sort of thing is important for readable and maintainable code.

mholt: (I know I haven't been very active with maintaining this repo lately, but) Yes, asking for formatting improvements is a normal part of most pull requests in Go projects -- in fact, a lot of CI are configured to fail if `go fmt` is not run on the code first. Thanks for understanding!

PR #265, opened 2020-11-18
apply correct timestamps to files

### LINKED ISSUE #212: Tar extraction should set tar-level timestamps
When extracting a tar (or tar composition, i.e., tar.xz, tar.gz, tar.bz2, etc), I would like the extracted files to have the proper timestamps (atime, mtime) that were recorded in the file

This is the behavior of `GNU tar`, and lacking this feature breaks some usecases

I can't use a tar of bazel's cache via `generic-worker` project (https://github.com/taskcluster/taskcluster/blob/222e8057550e7560412920446e76076d3db8e496/workers/generic-worker/mounts.go#L670-L679): newer timestamps breaks sanity checks from Bazel.

illiliti:
Closes #212

### DIFF
--- a/rar.go
+++ b/rar.go
@@ -197,33 +197,30 @@ func (r *Rar) unrarFile(f File, to string) error {
 		return fmt.Errorf("expected header to be *rardecode.FileHeader but was %T", f.Header)
 	}
 
-	if f.IsDir() {
+	if err := mkdir(filepath.Dir(to), 0755); err != nil {
+		return err
+	}
+
+	var err error
+
+	switch {
+	case f.IsDir():
 		if fileExists("testdata") {
-			err := os.Chmod(to, hdr.Mode())
-			if err != nil {
-				return fmt.Errorf("changing dir mode: %v", err)
-			}
+			err = os.Chmod(to, hdr.Mode())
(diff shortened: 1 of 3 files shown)


illiliti:
This patch is highly needed for building release tarballs with pre-generated `./configure` script.
Without this patch you will get this error:
```
WARNING: 'automake-1.16' is missing on your system.
         You should only need it if you modified 'Makefile.am' or
         'configure.ac' or m4 files included by 'configure.ac'.
         The 'automake' program is part of the GNU Automake package:
         <https://www.gnu.org/software/automake>
         It also requires GNU Autoconf, GNU m4 and Perl in order to run:
         <https://www.gnu.org/software/autoconf>
         <https://www.gnu.org/software/m4/>
         <https://www.perl.org/>
make: *** [Makefile:700: Makefile.in] Error 1
```

Jacalz:
I'll see if I can review this sometime in the coming weeks. Have a bit much right now.

mholt:
Thanks for this. I think it is made irrelevant with #302, but when I reimplement the `arc` command I'll try to remember to preserve timestamps in the extracted files.

PR #269, opened 2020-12-01
Check if symlink points outside of destination directory

### LINKED ISSUE #242: [sec] check if symlink points outside of destination directory
I was able to confirm that `os.Create()` will happily follow symlinks.

Although standard tools like `tar` will not likely allow you to add two files with the same name to an archive file, it's certainly easy enough to do in code. This means that **we need to add an additional check** before writing to a destination to make sure that it's not a symlink outside of the destination.

The vulnerability only exists if `OverwriteExisting` is _also_ turned on.

## What file is affected?

`filecompressor.go`:
```go
// DecompressFile reads the source file and decompresses it to destination.
func (fc FileCompressor) DecompressFile(source, destination string) error {
    if fc.Decompressor == nil {
        return fmt.Errorf("no decompressor specified")
    }

    if !fc.OverwriteExisting && fileExists(destination) {
        return fmt.Errorf("file exists: %s", destination)
    }

    // TODO needs check here

    in, err := os.Open(source)
    if err != nil {
        return err
    }
    defer in.Close()

    out, err := os.Create(destination)
    if err != nil {
        return err
    }
    defer out.Close()

    return fc.Decompress(in, out)
}
```

## Possible solution

If the destination exists we should always check if it's a symlink first:

```go
    if fileExists(destination) {
        // TODO check if destination is a symlink with an outside target
        // should be a simple os.Lstat()

        if !fc.OverwriteExisting {
            return fmt.Errorf("file exists: %s", destination)
        }
    }
```

Also, we could block ALL symilnks that have a target outside of the destination directory.

I think the former approach is the best for now, as some archives may link outside of the destination directory for legitimate reasons.

## Please link to any related issues, pull requests, and/or discussion

This is related to 

- https://github.com/mholt/archiver/pull/203
- https://github.com/mholt/archiver/pull/169
- https://github.com/mholt/archiver/pull/70
- https://github.com/snyk/zip-slip-vulnerability/issues/41

tslocum:
Resolves #242.

### DIFF
--- a/filecompressor.go
+++ b/filecompressor.go
@@ -3,6 +3,7 @@ package archiver
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 )
 
 // FileCompressor can compress and decompress single files.
@@ -47,8 +48,23 @@ func (fc FileCompressor) DecompressFile(source, destination string) error {
 	if fc.Decompressor == nil {
 		return fmt.Errorf("no decompressor specified")
 	}
-	if !fc.OverwriteExisting && fileExists(destination) {
-		return fmt.Errorf("file exists: %s", destination)
+
+	if fileExists(destination) {
+		if !fc.OverwriteExisting {
+			return fmt.Errorf("file exists: %s", destination)
+		}
+		// Disallow symbolic link targets outside of the destination directory.
+		info, err := os.Lstat(destination)
+		if err != nil {
+			return fmt.Errorf("failed to lstat destination file %s: %s", destination, err)
+		} else if info.Mode()&os.ModeSymlink != 0 {
+			linkDest, err := os.Readlink(destination)
+			if err != nil {
+				return fmt.Errorf("failed to read symbolic link %s: %s", destination, err)
+			} else if !within(filepath.Dir(destination), linkDest) {
+				return fmt.Errorf("symbolic link %s has a target outside of the destination directory: %s", destination, linkDest)
+			}
+		}
 	}
 
 	in, err := os.Open(source)


Jacalz (approved):
This is looking good to me. Compiles and works fine from my testing.
Ping @mholt as it would be great to get this security improvement reviewed and merged soon-ish.

filecompressor.go
```diff
@@ -47,8 +48,23 @@ func (fc FileCompressor) DecompressFile(source, destination string) error {
...
+
+	if fileExists(destination) {
+		if !fc.OverwriteExisting {
+			return fmt.Errorf("file exists: %s", destination)
+		}
+		// Disallow symbolic link targets outside of the destination directory.
+		info, err := os.Lstat(destination)
+		if err != nil {
+			return fmt.Errorf("failed to lstat destination file %s: %s", destination, err)
+		} else if info.Mode()&os.ModeSymlink != 0 {
```
servusDei2018, line 60: There is no need for nesting this in another `else` block; if `err != nil`, we already return.

filecompressor.go
```diff
@@ -47,8 +48,23 @@ func (fc FileCompressor) DecompressFile(source, destination string) error {
...
+			linkDest, err := os.Readlink(destination)
+			if err != nil {
+				return fmt.Errorf("failed to read symbolic link %s: %s", destination, err)
+			} else if !within(filepath.Dir(destination), linkDest) {
```
servusDei2018, line 64: There is no need for nesting this in another `else` block; if `err != nil`, we already return.

servusDei2018 (approved):
Good work, although I would avoid repetitive nesting.

tslocum:
Seeing this PR get approved and then closed without any feedback is really discouraging.

mholt:
There was feedback. The PR was inactive. And v3 is no longer being maintained in favor of v4, which doesn't even write to disk (currently -- the command needs to be rewritten for v4).

PR #272, opened 2020-12-23
Get the dependencies updated for the latest fixes and improvements

Jacalz:
Every package has been updated where possible to incorporate the latest performance fixes, improvements and bug fixes.
Each individual package has been updated in separate commits for easy cherry picking.
One important update is the update of https://github.com/ulikunitz/xz to [v0.5.9](https://github.com/ulikunitz/xz/releases/tag/v0.5.9) that bring a fix for https://github.com/ulikunitz/xz/security/advisories/GHSA-25xm-hr59-7c27.

PR #273, opened 2020-12-23
Switch zip handling to use github.com/klauspost/compress/zip

Jacalz:
This improves the performance over the stdlib implementation.
Archiving seems to be about 2.5x faster in my tests and archiving seems to be around 1.2x faster.
Those numbers should be even better after https://github.com/mholt/archiver/pull/272 is merged with the updated of the compression library.

### DIFF
--- a/cmd/arc/main.go
+++ b/cmd/arc/main.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"archive/tar"
-	"archive/zip"
 	"bytes"
 	"compress/flate"
 	"flag"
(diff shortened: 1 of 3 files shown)


mholt:
Thanks, this is a good idea. I think when I wrote this package, the upstream wasn't quite there yet. But I think we use it in Caddy now too so I approve.

Make sure go.mod/go.sum are updated accordingly if necessary!

Jacalz:
This worked good even without the updating `github.com/klauspost`, so I just figured that I would separate the two. This should be ready as is :)

mholt (approved):

Issue #264, opened 2020-11-18
How to set default options?


danez:
I would like to set `StripComponents = 1` for all types which support it and use `archiver.Unarchive()`.

I cannot figure out how to change the default options though. First I tried doing `archiver.DefaultTar.StripComponents = 1`, but I see now that `Unarchive()` is always creating a new instance.

Is there a way to do it?

Or-Geva:
I have the same issue with `OverwriteExisting` which by default is false.

mholt:
This can be done with the v4 rewrite introduced in #302.

Issue #266, opened 2020-11-20
Creat new file (or folder) if it already exists, instead of just returning an error


Jacalz:
## What would you like to have changed?

Currently, if the path to extract the zip file to already exists, the method will return an error. It is common in other software to add a number at the end to the filename and save it to the side of the save path. For example, saving the file for the first time it already exists, the file will be `filename(1)`, the next time `filename(2)` and so on.

The method being used:
https://github.com/Jacalz/wormhole-gui/blob/449c464f9096d33c4d4c63b78a70f4ba9509f0e9/internal/bridge/receiver.go#L83

## Why is this feature a useful, necessary, and/or important addition to this project?

It avoids having each user of the API implementing this logic themselves and makes the API a lot more useful. I would like to see this implemented for https://github.com/Jacalz/wormhole-gui to make it much easier to save a file where one with the same name already exists.

## What alternatives are there, or what are you doing in the meantime to work around the lack of this feature?

I would need to check the error and append `(number)` to the end of the file name and try multiple times. This is not only slow, but impractical to do multiple times.

Jacalz:
My notes about this is specific to the zip unpacking. I don't know if it is an issue with other formats.

Jacalz:
This can easily enough be done by the developer and I think that this mostly would complicate the API. Going to close this as I have no interest in this specific feature as it stands.

PR #274, opened 2021-01-03
Customize default values

Or-Geva:
This PR adds the ability to set the default values for all types which support it, e.g.:
Right now the following code will use the default values of a gz archiver instance. In order to set the default values (e.g. override=true), we have to create an instance of gz and configure it.

```err = archiver.Unarchive("test.tar.gz", "test")```

After this PR will be merged, you will be able to create a new archiver instance and configure it before use e.g.:
```
func Unarchive(localArchivePath, originArchiveName, destinationPath string) error {
	// Create archiver instance based on file extention.
	archiverInterface, err := archiver.ByExtension(originArchiveName)

	// Create default values.
	params := archiver.CustomizeParams{
		Tar: &archiver.Tar{
			OverwriteExisting: true,
		},
	}

	// Set default values for an archiver instance.
	archiverInterface, err = archiver.Customize(archiverInterface, params)
	if err != nil {
		return err
	}
	u, ok := archiverInterface.(archiver.Unarchiver)
	if !ok {
		return errorutils.CheckError(errors.New("format specified by source filename is not an archive format: " + originArchiveName))
	}

	// Run the archiver methods.
	return u.Unarchive(localArchivePath, destinationPath)
}
```
Fixes: https://github.com/mholt/archiver/issues/264

### DIFF
--- a/archiver.go
+++ b/archiver.go
@@ -480,6 +480,66 @@ func ByExtension(filename string) (interface{}, error) {
 	return nil, fmt.Errorf("format unrecognized by filename: %s", filename)
 }
 
+// Customize customize an archiver and unarchiver, or compressor
+// and decompressor, based on CustomizeParams struct.
+func Customize(f interface{}, params CustomizeParams) (interface{}, error) {
+	switch v := f.(type) {
+	case *Rar:
+		v.OverwriteExisting = params.OverwriteExisting
+		v.MkdirAll = params.MkdirAll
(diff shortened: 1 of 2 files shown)


Jacalz:
I don't see how this mentions closing #266. I think you must have typed the wrong issue.

Or-Geva:
@Jacalz, you are right, thanks.

PR #277, opened 2021-01-12
Use go error wrapping on header writing for easier error handling

vansante:
Followup on #276 

I now get errors like these:

`path/to/file: making header: write tcp 10.2.0.6:8080->10.99.12.11:52284: write: broken pipe`

I would like to filter these too :)


mholt (approved):
Sounds good, thanks for the use case!

Issue #279, opened 2021-02-16
Exclude files by pattern


atsikham:
## What would you like to have changed?

Currently there is only an ability (at least it's documented so) to specify files/directories to add into archive. Exclude some files from specified directories by pattern would be helpful.

## Why is this feature a useful, necessary, and/or important addition to this project?

My use-case is filtering logs with *.log extension out, but I'm sure there are a lot of other ones.

## What alternatives are there, or what are you doing in the meantime to work around the lack of this feature?

Make a copy of the directory to compress, remove unnecessary files, compress.

nononsensecode:
There is also one use-case, when we want to exclude the .git directory

mholt:
We can probably find a way to work this into #302, which is a complete rewrite of this package and destined to become v4.

mholt:
One option in v4 is to use `FilesFromDisk()` and then simply remove the files you don't want from the resulting slice. :man_shrugging: 

We can reopen this if needed, but I think that'll work fine unless there's a use case not covered sufficiently by that solution.

NotGabry:
is there any way to exclude the file directly from `FilesFromDisk()` beacause when (in linux) archiver tries to read a file that is locked by another program it crashes, an idea cloud be adding `false` to the path of file like
```golang
files, err := archiver.FilesFromDisk(nil, map[string]string{
	"/home/randuser/folder/randfile.txt.lck": false,
	"/home/randuser/folder": "folder"
})
```

mholt:
But `false` is not a string. And `"false"` could be a filename, so I don't think that will do.

> beacause when (in linux) archiver tries to read a file that is locked by another program it crashes

Is that a Windows-thing? What is the output? Archiver shouldn't _crash_ (panic) but if you get an error, that's understandable.

PR #281, opened 2021-03-24
Filter function for zip archiver

jhwz:
If need be a filter function can be added to the other archivers, it's pretty straight forward!

Need for similar reasons as described in #279


mholt:
Thanks for contributing this. I think #302 makes this irrelevant or even fixes this, since when creating an archive, you can build the list of files you want. `FilesFromDisk()` can do this for you, and if you want to skip any that it added, you can remove them from the list.

PR #284, opened 2021-04-20
Change hasTarHeader to hasRarHeader in rar.go

xuiqzy:
I think that was a typo, since the bytes in the byte slice even say "Rar", so I guess it is a rar specific header being checked here.


mholt (approved):
Yep, copy-pasta error.

PR #295, opened 2021-10-01
Update README.md

lpar:
Fix the install instructions for current versions of Go, since trying the current suggestion will likely lead the user through a trail of unhelpful error messages.


mholt (approved):
Forgot about this. Thanks!

Issue #282, opened 2021-04-19
Is it possible to unrar a file from memory rather than having to write it to disk and pass it a file name?


mcastrogg:
Hello, a bit new to go so apologies if i'm missing something obvious here. But is there a simple way to pass an ioReader containing a .rar to one of the functions and then init the Rar from the ioReader rather than having to pass it the file path? 

Thanks.

mholt:
There is now! That is what #302 does, which is a complete rewrite of this package and will become v4.

Issue #278, opened 2021-01-25
Remove unconditional extention-check in library code


fahman:
## What would you like to have changed?
In all of the implementations of `Archiver` a hard-coded `CheckExt` is being invoked within `Archive` func. I propose removing this check from `Archive` and `Unarchive` and instead move it to the code that e.g. determines `Archiver` to be used. The operations of specific `Archiver` should not do any extension checking.

## Why is this feature a useful, necessary, and/or important addition to this project?
When using archiver as library there are plenty of scenarios (e.g. temp files, in-memory files being compressed before transfer as a payload, number of well-known file names that internally are just archives, etc.) when file extension is just irrelevant. When selecting archiver implementation e.g. by `archiver.DefaultTar` the decision about format is already carried out so checking for extension within funcs of selected implementation seems to be redundant.

## What alternatives are there, or what are you doing in the meantime to work around the lack of this feature?
Current hack is to always create a temp file with the required extension and copy stuff back and forth, which causes boiler-plate code and additional ways to break code.

## Please link to any relevant issues, pull requests, or other discussions.
Similar request here: https://github.com/mholt/archiver/issues/255

mholt:
It's your lucky day, #302 does this. I completely redesigned this package and its core APIs don't rely on the file extension anymore (except for an optional `Identify()` function that can optionally use the file extension -- or the stream contents directly -- to match formats). That PR will become the basis for v4. Please try it out!

Issue #262, opened 2020-11-14
Absolute paths for files within tar archives


d-Rickyy-b:
## What would you like to have changed?

Ability to enable storing of files in tar archive with their absolute paths.

## Why is this feature a useful, necessary, and/or important addition to this project?

You might not want to archive every subdirectory within your crafted structure. So you might specify a more deeply nested directory but want to preserve higher directories (maybe even up to the file system root) for restoring the files later on.

An example:

Imagine you are storing files in /home/rico/project/ and within that directory you have the following structure:
```
project/
├── huge_files
├── images
├── src
└── tmp
```

Now I'd prefer to get an archive file with the contents of `src` and `images` and avoid adding the tmp and "huge" files to my archive, because that's not what I need in this case. On the other hand I still want to preserve the directory structure to know "where it belongs to". Maybe I have several projects all having a "src" folder. Knowing by the archive paths where which archive belongs to is quite helpful. The final archive would contain the following paths & files:

```
/home/rico/project/src/<files>
/home/rico/project/images/<files>
```

## What alternatives are there, or what are you doing in the meantime to work around the lack of this feature?

1. When possible using tar -cvf /absolute/path/to/file
2. Modifying your code to add a new check
-> https://github.com/mholt/archiver/blob/25e050dbb4e3c591402916201098241bb661a2f7/tar.go#L303-L317
I have added one line in line 311: 
```topLevelFolder = filepath.Dir(source)```
That way we always get the full path.

## Please link to any relevant issues, pull requests, or other discussions.

#250

coolaj86:
For the record: I actually agree with you, personally.

But as per the discussion in #250, we probably won't change the behavior.

However, I would like to know why you want to use `arc` rather than `tar` (or Go's native tar package) for this use case?

d-Rickyy-b:
Thanks for your comment. One reason why I want to use archiver is because it provides a **way** better API than the native tar package.
Another reason is, that I also wanted zip support. And archiver abstracts the different archive formats away. One package for multiple formats.

Actually I went with using the tar package for now, because it gives me the freedom to use absolute paths - but I would be way happier with the archiver API.

mholt:
This is fixed in #302, where you specify the mapping of file paths on disk to the file paths in the archive when you create the archive. Please check it out! :) We can reopen this issue if it is still unresolved or needs more discussion.

Issue #255, opened 2020-10-20
Option to ignore file extension and detect archive format by file content


yene:
## What would you like to have changed?
An option to ignore the file extension and detect the archive format based on the file content. 

## Why is this feature a useful, necessary, and/or important addition to this project?
Some of our customers deliver file formats with wrong extensions.

## What alternatives are there, or what are you doing in the meantime to work around the lack of this feature?
No idea. Maybe a bash file -I to check the type.

coolaj86:
There's now a `--ext` option which you can use set set the type explicitly, as per https://github.com/mholt/archiver/pull/252, released with v3.4.0.

I don't think that we're going to do any type of sniffing on files, but you could do something like this, if it's a script environment:

```bash
arc -ext rar unarchive ~/Downloads/bat-v0.15.4-x86_64-apple-darwin.tar.gz
arc -ext tar unarchive ~/Downloads/bat-v0.15.4-x86_64-apple-darwin.tar.gz
arc -ext tar.bz2 unarchive ~/Downloads/bat-v0.15.4-x86_64-apple-darwin.tar.gz
arc -ext tar.gz unarchive ~/Downloads/bat-v0.15.4-x86_64-apple-darwin.tar.gz
arc -ext tar.xz unarchive ~/Downloads/bat-v0.15.4-x86_64-apple-darwin.tar.gz
arc -ext zip unarchive ~/Downloads/bat-v0.15.4-x86_64-apple-darwin.tar.gz
```

The best way to do that would be to try each

coolaj86:
You could also use `file` to help determine the file type by "magic headers":

```bash
file /tmp/evil/double-evil.rar
/tmp/evil/double-evil.rar: POSIX tar archive
```

coolaj86:
@yene Are the proposed workarounds sufficient to close the issue?

mholt:
This has been implemented for all formats in #302, which will become v4 soon. Please give it a try and let me know your thoughts!

PR #302, opened 2022-01-02
v4 rewrite: All new design, core types, and stream-oriented interfaces

### LINKED ISSUE #118: Add ByHeader support to all formats

mholt:
I spent the holiday rewriting this package from scratch with a completely new approach to handling archives and compression formats. This will become v4.

The new core APIs are completely stream-oriented and file-agnostic. The abstractions for files, directories, and file systems are virtualized thanks to the recently-added `io/fs` package in the Go standard library. I expect these design changes will close most open issues and PRs because it either fixes the problems or makes them irrelevant.

A significant number of issues relate directly to interactions with specific file systems (files on disk) and this new API does not deal with that directly anymore, except for a couple specific functions that read the disk in order to create the abstraction. Nothing in the core API writes to disk or deals with that. (Yay!) The stream and FS abstractions are highly flexible to build upon.

Another nice thing about the new design is that there's no more need for explicit composite types (like `TarGz` and `TarBz2`), because we have a new `CompressedArchive` type that composes an archive and a compression format. It's mainly used with Tar only, since Zip and Rar do their own thing, but another nice feature is the `Identify()` function that automatically gets you the right type:

```go
// opening a file on disk for this example, but can be any ReadSeeker
unknownFile, err := os.Open("filename.tar.gz")
if err != nil {
	return err
}
defer unknownFile.Close()

// if you don't have a filename, leave it blank: identification uses both/either filenames and/or streams
format, err := archiver.Identify("filename.tar.gz", unknownFile)
if err != nil {
	return err
}

// we can now work with the file, for example, extract a file out of it
if ex, ok := format.(archiver.Extractor); ok {
	ex.Extract(context.Background(), unknownFile, "target.txt", func(_ context.Context, f File) error {
		// do something with the file ...
		return nil
	})
}

// or maybe it's just a compressed log file
if decom, ok := format.(archiver.Decompressor); ok {
	rc, err := decom.OpenReader(unknownFile)
	if err != nil {
		return err
	}
	defer rc.Close()
	// read from it ... all reads are decompressed now
}
```

I think the new APIs are pretty slick, and you should try them out and let me know what you think.

**One of my favorite new features is the `FileSystem()` function. Give it a path on disk, and it will return a `fs.ReadDirFS`. Basically, it lets you read from real directories, regular files, archive files, and compressed archive files, ALL THE SAME WAY.** This is pretty cool I think. You don't have to worry about whether the given file is just a regular file, a directory, or an archive file (which _acts_ like a directory because it contains other files!) -- you can traverse it all the same way. And the archive format doesn't matter either, it's automatically identified for you! You literally won't even know things are being decompressed as you read from them:

```go
fsys, err := archiver.FileSystem("/path/to/folder/or/file")
if err != nil {
	return err
}

// traverse everything except the ".git" folder...
err = fs.WalkDir(fsys, ".", func(path string, d fs.DirEntry, err error) error {
	if err != nil {
		return err
	}
	if path == ".git" {
		return fs.SkipDir
	}
	fmt.Println(path, d.IsDir())
	return nil
})
if err != nil {
	return err
}

// ...or just open one file
file, err := fsys.Open("example.txt")
if err != nil {
	return err
}
defer file.Close()
```

To make a new archive from files on disk, you might do this:

```go
files, err := archiver.FilesFromDisk(map[string]string{
	"/path/on/disk/file1.txt": "file1.txt",
	"/path/on/disk/file2.txt": "subfolder/file2.txt",
	"/path/on/disk/folder":    "",
})
if err != nil {
	return err
}

out, err := os.Create("example.tar.gz")
if err != nil {
	return err
}
defer out.Close()

caf := archiver.CompressedArchive{
	Compression: archiver.Gz{},
	Archival:    archiver.Tar{},
}

err = caf.Archive(context.Background(), out, files)
if err != nil {
	return err
}
```

Notice how you have the flexibility of mapping each file (or folder) to a different path in the archive. You can also leave the mapped path blank to assume the base filename for convenience. Folders are added recursively.

Oh yeah, and I added basic `context` support.

The `arc` command has not been ported over yet, and as that has to deal directly with the file system, it will require some work before it is ready. This would be the only part of the repository that would write to disk, as the core library doesn't write to disk anymore.

Looks like I was able to delete about 70% of the code, however that count includes test files, the command, and the README, which have not been restored yet. Still, I could feel a significant code reduction in this new design when I wrote it.

Should make irrelevant / close / fix #118, #128, #131, #141, #146, #150, #227, #194, #204, #216, #239, #255, #262, #278, #282


mholt:
I might rename this package to `archives`. It was suggested before but I didn't get around to it. Maybe now is a good time.

coolaj86:
> I might rename this package to `archives`. It was suggested before but I didn't get around to it. Maybe now is a good time.

I believe in the Go philosophy that if you're making a breaking change, that's a version bump, but if you're making a philosophical change, that's a new package.

coolaj86:
This isn't a PR that one can simply review. This is a new project from the ground up.

Make a new repo, get some tests in, mark this as deprecated, and link to the new repo.

How do you want to handle tests in the new project?

mholt:
Ha, sorry @coolaj86 , I tagged you as a reviewer not with the expectation that you'd actually give a full review -- I probably should have just @-mentioned you instead, since all I wanted to do was let you know that this is inbound. :)

mholt:
After quite a few more trials, this seems to be working decently well at least, so I am going to merge it into master but I won't tag it yet (except pre-releases).

Some more tests would be nice, as well as an actual command, but for now I want to have some people start using this, and I need to move my focus onto some other projects here pretty soon anyway.

The new README shows off some of the new version's elegance and features, so I hope you'll enjoy!

jeff-mccoy:
@mholt this is a substantial update and a lot of really great additions.  I'm very interested in the stream options now as we have a new use for them for our air gap delivery tool that uses your library.  Thanks for the work on this, I see some notes about needing testing/validation, but also notice the tests seem to be less robust in this new cut.  I don't think we'll be able to jump to v4 without some more automated tests, but will play with this new version for sure.  I also saw you have other priorities soon, do you have an idea what the plan will be for automated tests with v4 going forward?  

On another note, I spent the holiday break doing a "mega PR" myself for our project that caused some excitement 🤣:  https://github.com/defenseunicorns/zarf/pull/237.

mholt:
@jeff-mccoy That's great to hear! I agree v4 could benefit from more tests.

>  I also saw you have other priorities soon, do you have an idea what the plan will be for automated tests with v4 going forward?

Yeah, so that's true... I'm really busy with Caddy/CertMagic right now, and what I have written for archiver v4 works for me and my needs. However, I do accept [sponsorships](https://github.com/sponsors/mholt), and at the right tiers I can absolutely prioritize writing some automated tests for you. :+1:

Otherwise, I can try to get around to them when I have a chance, or I'm happy to accept contributions (PRs) as well! Let me know what suits you best and we can do that.

jeff-mccoy:
Awesome.  I think we'll do some combination of those options.  We deeply value open source projects and know it can be a thankless endeavor sometimes.  We'll keep in touch for sure, thanks again for the quick reply and the great library.

mholt:
@jeff-mccoy Sounds good -- would look forward to working with you on this if that can happen!

mholt:
@jeff-mccoy Just checking in, how is this lib working for you? (It's been through several releases now that enhance its functionality and fix bugs.)

Issue #303, opened 2022-01-08
Dereference symlinks flag for FilesFromDisk


korpa:
## What would you like to have changed?

Instead of adding the symlink into archive, it should add the content of the target file of the symlink if the flag is set. Similar to https://www.gnu.org/software/tar/manual/html_node/dereference.html

mholt:
Thanks for the request. I don't think anything needs to change about `Archive()`'s signature to do what you want. All `Archive()` does is create an archive with the files in the list. To dereference symlinks, don't create a `File` with a `FileInfo` that sets the SymLink flag and has `LinkTarget` set. Instead, create a `File` with an `Open` func set and `LinkTarget` unset/empty.

Does that make sense? So basically you'll do the dereferencing the way you want, instead of leaving that up to the archiver package.

(Closing, but feel free to continue discussion. We can reopen if needed.)

korpa:
I'm not sure if I understand what you mean. Currently I use `archiver.FilesFromDisk` so I haven't touched any functions you mention, yet. I will have a look at at your proposal and try to understand.

On the other hand, maybe it would make sense to add a dereference flag to `FilesFromDisk`.

korpa:
I had a look at `FilesFromDisk` and now got what you mean.

I could implement my own `FilesFromDisk`, but I think a dereference option would be a good fit in the official `FilesFromDisk` function. 
What do you think?

mholt:
Ah, yeah, that's a totally separate function from what was mentioned originally.

Maybe it would make sense for that function to take options. I'm mobile right now but we could discuss a few ideas. 👍

korpa:
I added a pull request but I'm not sure if it is the best solution. Let's discuss.

PR #304, opened 2022-01-08
Add dereference option to FilesFromDisk

korpa:
This is a first suggestion how this feature could look like.

Solves #303

### DIFF
--- a/archiver.go
+++ b/archiver.go
@@ -46,35 +46,13 @@ func (f File) Stat() (fs.FileInfo, error) { return f.FileInfo, nil }
 // their associated names in the archive. For convenience, empty values are
 // interpreted as the base name of the file (sans path) in the root of the
 // archive. Keys that specify directories on disk will be walked and added
-// to the archive recursively, rooted at the named directory. Symbolic links
-// will be preserved.
+// to the archive recursively, rooted at the named directory.
+//
+// File gathering will adhere to the settings specified in options.
 //
 // This function is primarily used when preparing a list of files to add to
 // an archive.
-
-func FilesFromDisk(filenames map[string]string) ([]File, error) {
-	files, err := filesFromDiskGeneric(filenames, false)
-	return files, err
-}
-
-// FilesFromDiskDerefSymlinks returns a list of files by walking the directories in the
-// given filenames map. The keys are the names on disk, and the values are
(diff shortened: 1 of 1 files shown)


mholt (changes requested):
Nice, thanks for the implementation.

I think the API for this could be improved. How about if we keep just the one function, `FilesFromDisk()`, where the first argument is a struct of options for assembling the list of files? The only option for now would be whether to follow symlinks.

korpa:
I had the same idea, but I didn't want to change parameters for `FilesFromDisk()` to not break the function.

mholt:
I wouldn't worry about that. V4 is in alpha so it is still under active development. Breaking changes are allowed and expected, I've already broken the v4 API a few times this week.

korpa:
I will have no time until the weekend for this. So please go ahead. If you haven't done it till Saturday, I will jump in.

mholt:
@korpa No worries. I just pushed my version of the change. What do you think? Let me know if you have a chance to try it.

mholt:
I think I'll probably merge this today and tag a new alpha version, if you have a chance to check it out you can!

korpa:
Please let me know, as soon as you have the new alpha out.

mholt:
@korpa Will do. Do you want to test this first to make sure it does what you want?

korpa:
Looks very good. I replaced my personal "filesFromDisk" function with `files, err := archiver.FilesFromDisk(archiver.FromDiskOptions{FollowSymlinks: true}, filesDirs)`, created my tar.gz and extracted it

=> Instead of a symlink there was a file with the content as expected.

korpa:
Before creating a new alpha should we add some info in the README?

mholt (approved):
Awesome, thank you for testing!!

I will merge this, update the README, and tag a new alpha release.

mholt:
@korpa Here is the tag, and I updated the readme: https://github.com/mholt/archiver/releases/tag/v4.0.0-alpha.3

I also made the options parameter optional (by making it a pointer). See the godoc :+1: Thanks again!

korpa:
alpha3 works as expected as well after changing the options parameter to a pointer. 

Thank you very much for adding this feature.

mholt:
Great! And thank you for the initial PR.

PR #305, opened 2022-01-10
zip: add auto encoding in FileSystem

pashifika:
archiver.FileSystem is not support zip TextEncoding settings.

### DIFF
--- a/zip.go
+++ b/zip.go
@@ -24,7 +24,7 @@ import (
 )
 
 func init() {
-	RegisterFormat(Zip{})
+	RegisterFormat(Zip{TextAutoEncoding: true})
 
 	// TODO: What about custom flate levels too
 	zip.RegisterCompressor(ZipMethodBzip2, func(out io.Writer) (io.WriteCloser, error) {
@@ -78,6 +78,22 @@ type Zip struct {
 	// encoded filenames and comments, specify the character
 	// encoding here.
 	TextEncoding string
+
+	/*You must set AutoEncodings to use it, this is fix to FileSystem decode error.
+	Example:
+
+	import (
+	    "github.com/mholt/archiver/v4"
+	    "golang.org/x/text/encoding"
+	    "golang.org/x/text/encoding/japanese"
+	    "golang.org/x/text/encoding/simplifiedchinese"
+	)
+	archiver.AutoEncodings = []encoding.Encoding{
+	    japanese.ShiftJIS,
+	    simplifiedchinese.GB18030, // set last at any time
+	}
+	*/
+	TextAutoEncoding bool
 }
 
 func (z Zip) Name() string { return ".zip" }
@@ -213,7 +229,7 @@ func (z Zip) Extract(ctx context.Context, sourceArchive io.Reader, pathsInArchiv
 // It is a no-op if the text is already UTF-8 encoded or if z.TextEncoding
 // is not specified.
 func (z Zip) decodeText(hdr *zip.FileHeader) {
-	if hdr.NonUTF8 && z.TextEncoding != "" {
+	if hdr.NonUTF8 && (z.TextEncoding != "" || z.TextAutoEncoding) {
 		filename, err := decodeText(hdr.Name, z.TextEncoding)
 		if err == nil {
 			hdr.Name = filename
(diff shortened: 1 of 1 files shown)


mholt:
Thanks for the PR. I see what you are trying to do with this, I wonder if we can make it a little more elegant. I will try to circle back on this soon with some feedback. Thank you

pashifika:
@mholt 
Thanks, rar files also have this problem (older versions).

mholt:
@pashifika What if we simply automatically try all of the known encodings (if enabled)? Rather than having to register them. This way you just have to set `TextAutoEncoding` to true and it will just try them all. Maybe it can remember the last successful encoding and try it first, for efficiency. I think that is a cleaner API.

mholt:
I could probably post my proposed change pretty quick. Do you want to try it or should I?

pashifika:
@mholt 
thanks. I tried it at first, but I think it's better not to record it because there are various inconveniences.

pashifika:
```
/123
/123/abc.txt
/あいうえお/
/あいうえお/あいうえお.txt
```
If you save it, it will be a bug in a ZIP file with such a structure, I think that the correct solution is probably to check the language of the system and set it in advance (or throw it to the user and set the list from AutoEncodings matter?

mholt:
Would this package be helpful? https://github.com/jeandeaual/go-locale

pashifika:
I think that's a good idea.

pashifika:
```go
type charset []encoding.Encoding

func (c *charset) UnmarshalFlag(value string) error {
	var err error
	if value == "" {
		err = errors.New("charset is empty")
	} else {
		strList := strings.Split(strings.ToLower(value), ",")
		var charList []encoding.Encoding
		checkMap := map[string]bool{}
		for _, str := range strList {
			if _, ok := checkMap[str]; ok {
				// Duplicate deletion process
				continue
			}
			switch str {
			case "ja", "ja-jp":
				charList = append(charList, japanese.ShiftJIS)
			case "zh", "zh-cn":
				charList = append(charList, simplifiedchinese.GB18030)
			case "ko", "ko-kr":
				charList = append(charList, korean.EUCKR)
			default:
				err = fmt.Errorf("%s charset is not supported", str)
				break
			}
			checkMap[str] = true
		}
		if err == nil {
			*c = charList
		}
	}
	return err
}
```

As an aside, this is the code I'm currently using

mholt:
@pashifika Do you have a sample archive file I could test with? That was really helpful on the last issue.

pashifika:
@mholt 
[あいうえお・テスト卵.zip](https://github.com/mholt/archiver/files/7868464/default.zip)
[あいうえお・テスト卵.rar.zip](https://github.com/mholt/archiver/files/7868465/rar.zip)
[あいうえお・テスト卵.7z.zip](https://github.com/mholt/archiver/files/7868470/7z.zip)

mholt:
(lol, テスト卵 -- from the little bit of Japanese I know I appreciate the funny filename)

pashifika:
@mholt 
can you see this commit?
https://github.com/pashifika/archiver/commit/c50b17fe040ccee268474eca21da3c354d9d4451

mholt:
@pashifika Yeah. Is that fixing a bug?

pashifika:
@mholt 
good night.
yes, during file compression, all files in the directory are compressed to the root directory.

mholt:
@pashifika Come to think of it, I don't know if I tested subfolders with .zip files when I was developing it. So that is a very good point. Can you create a pull request? Maybe I can merge it in before I fall asleep.

:sleeping:

pashifika:
@mholt 
I'm off duty, maybe tomorrow.

mholt:
No worries. In fact, it's a simple fix, I just went ahead and did it here: f9ab7f5c4a302b4d9e86dfddbf4ee56f03d2b9d2

Cheers. Will continue looking at text encoding over the next few days. Turns out the bug I mentioned a few minutes earlier is fixed when the text encoding is set properly.

mholt:
Ok, after fiddling with this all morning, I was not able to find a reliable, consistent way to accurately detect charset automatically.

So I am probably not going to merge this change. However, I did include the fix for decoding zip filenames before checking fileIsInluded, in f9ab7f5. Thanks for that :+1:

I think the best thing to do if needing an alternate encoding with a file system is to manually set it:

```go
if afs, ok := fsys.(archiver.ArchiveFS); ok {
		if zip, ok := afs.Format.(archiver.Zip); ok {
			zip.TextEncoding = "shiftjis"
			afs.Format = zip
			fsys = afs
		}
	}
```

This worked for me most reliably. If a better, more reliable way is presented that can do it automatically, feel free to suggest it; I'm open to it. But it turns out that just "guessing and checking" isn't a good strategy here I think.

Issue #307, opened 2022-01-11
v4: 


tgulacsi:
Please include lines https://github.com/mholt/archiver/blob/10c5080fa78f78d10e28abdf11fa6d4abb7f999f/fs.go#L40 - https://github.com/mholt/archiver/blob/10c5080fa78f78d10e28abdf11fa6d4abb7f999f/fs.go#L53 as an example for library use (handle archive as an io.FS) - it took me an hour to find out how to do it, and it is essentially VERY easy, but this information was hard to find.

mholt:
Good question, so you're basically looking for an example of how to use `Identify()` -- isn't that basically what is already here? https://github.com/mholt/archiver#identifying-formats

tgulacsi:
Ok, I revise my thinking, and now I don't see an easy way to create an ArchiveFS for an in-memory io.ReadSeeker.
Which is strange, as ArchiveFS have everything, just needs a File member, and use that if not empty, and work as now if it is.
It has a lot of nice logic, that would be a shame to have duplicated several times...

mholt:
A file system over a stream? Like [I asked on Twitter](https://twitter.com/mholt6/status/1480271696338698241), I'm not sure how that would work. Can you explain how it would work and what it should do?

mholt:
Couple of random thoughts...

- The zip format can't use io.Reader, it needs io.ReaderAt, which is not part of the io.ReadSeeker interface. We already kind of hedge this by returning an error from Extract() if the input is not an io.ReaderAt and an io.Seeker. The interfaces are, unfortunately, disjoint.
- I could maybe see a FS that works on a stream only if all the operations are mutex'ed - can't seek to one place for one operation, then seek to another for another operation, at the same time. Currently ArchiveFS doesn't have this limitation because it opens multiple read-only handles to the files, one for each operation.

tgulacsi:
I have an io.SectionReader (over []byte of memory (maybe mmap'd)), which is an io.ReadSeekCloser and can create a SectionReader of it easily.

Maybe a `GetReader func() (io.ReadSeekCloser, error)` (like net/http.Request.GetBody) that is just `os.Open(a.Path)` if nil ?

mholt:
Hmm, I'm not sure if that gets us around the mutex requirements; and we might as well just have a field on the struct that lets you specify the stream directly rather than a path on disk from which to create the stream for you.

Will think on this and maybe try some things when I have a chance, and get back about it.

mholt:
[...]

PR #312, opened 2022-01-20
Add ExampleEmbed for ArchiveFS.Stream usage

tgulacsi:
For #307 (https://github.com/mholt/archiver/issues/307#issuecomment-1015110919): works very nice!

As a side note: can't happen that http.Dir is already what your DirFS tries to achieve?
net/http is a huge dependency, I wouldn't want to replace it, maybe just borrow the code, or look at the handling of some edge cases...

### DIFF
--- a/fs_test.go
+++ b/fs_test.go
@@ -1,6 +1,16 @@
 package archiver
 
-import "testing"
+import (
+	"bytes"
+	_ "embed"
+	"fmt"
+	"io"
+	"io/fs"
+	"log"
+	"net/http"
+	"path"
+	"testing"
+)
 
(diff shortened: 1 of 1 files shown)


mholt:
> As a side note: can't happen that http.Dir is already what your DirFS tries to achieve?

Nah, our DirFS does achieve what it needs to; just seems odd that os.DirFS doesn't implement ReadDir. I don't think we have a need for http.Dir.

Thanks for the example, but it doesn't compile and pass CI (missing resources).

mholt (approved):
Cool, thanks

PR #316, opened 2022-01-21
ClearAttributes option (close #280)

### LINKED ISSUE #280: Is there a way NOT to preserve file attributes?
Hi, thanks for creating this tool!

I use it for database archiving, i.e. I generate SQL of a database and then archive it. In case there were no changes in the DB, the content of those files is the same (md5 hash is the same), but the creation, modification, and access time stamps are different. When I use archiver to create an archive, the archive checksum is different of course because of those extended attributes of files that are within it.

Is there a way to disable preserving those attributes while creating archives? Obviously, all 3 would be set to the moment of extraction of such archive.

mholt:
@ivanduka @carolynvs @vdice Can you please test to make sure this does what you require?

### DIFF
--- a/archiver.go
+++ b/archiver.go
@@ -123,9 +123,9 @@ func FilesFromDisk(options *FromDiskOptions, filenames map[string]string) ([]Fil
 // noAttrFileInfo is used to zero out file attributes (issue #280)
 type noAttrFileInfo struct{ fs.FileInfo }
 
-func (noAttrFileInfo) Mode() fs.FileMode  { return 0 }
-func (noAttrFileInfo) ModTime() time.Time { return time.Time{} }
-func (noAttrFileInfo) Sys() interface{}   { return nil }
+func (no noAttrFileInfo) Mode() fs.FileMode { return no.FileInfo.Mode() | fs.ModeType }
+func (noAttrFileInfo) ModTime() time.Time   { return time.Time{} }
+func (noAttrFileInfo) Sys() interface{}     { return nil }
 
 // FromDiskOptions specifies various options for gathering files from disk.
 type FromDiskOptions struct {
--- a/archiver.go
+++ b/archiver.go
@@ -123,7 +123,7 @@ func FilesFromDisk(options *FromDiskOptions, filenames map[string]string) ([]Fil
 // noAttrFileInfo is used to zero out file attributes (issue #280)
 type noAttrFileInfo struct{ fs.FileInfo }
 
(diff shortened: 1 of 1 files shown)


carolynvs:
@mholt I tried it out but all of my files in the archive are empty. The headers are there, but no content.

There is a check during archival and all of my files are skipped from being copied into the archive because they aren't being recognized as regular files. I think the no file attr struct may need to either return more accurate info on the file mode?

mholt:
@carolynvs Thank you for trying it! Are you sure you used the latest commit on this branch? fbe2de8a40f53651bdde383c8030a5825fc0803e - it should preserve the file type of the mode, at least.

If you're using `go.mod`, you'd want to run `go get github.com/mholt/archiver/v4@fbe2de8a40f53651bdde383c8030a5825fc0803e` to get the expected version.

carolynvs:
If it helps here is the branch where I tried this out: https://github.com/carolynvs/porter/blob/8bc69d7775b8165ec8d0ea7e70fddbf1869e129e/pkg/porter/archive.go#L126-L151

carolynvs:
I just confirmed that I'm using the latest commit 👍

mholt:
Hmm you're right. That's strange. Am I not preserving the right bits? :thinking:

mholt:
@carolynvs D'oh, I'm an idiot -- `|` sets all the bits, `&` checks them. Please try the latest commit now :sweat_smile:

carolynvs:
Okay got a bit farther this time with the latest commit. Now the files are in the archive but the file permissions are 000, so I have to reset the read/execute bits in order to access the files after extracting.

My project was interested in clearing some attributes, like the timestamps, so that when we re-created an archive with the same file contents, it would generate an archive that was exactly the same as the original when hashed. I do need to preserve the file permissions.

Is there a way for me to do that with what's in this branch or would that require a change in archiver itself?

mholt:
@carolynvs Ah, okay -- thanks for being specific. I've just pushed a commit that also preserves the permission bits. Please give that a try and let me know!

carolynvs:
I got my test to pass by changing how we apply the bit mask a bit (using | to comine the two masks into one, then masking with &)

```go
// Mode preserves only the type and permission bits.
func (no noAttrFileInfo) Mode() fs.FileMode {
	rawMode := no.FileInfo.Mode()
	preservePerms := fs.ModePerm
	preserveFileType := fs.ModeType
	mask := preservePerms | preserveFileType
	return rawMode & mask
}
```

carolynvs:
Thanks for adding this! It'll be great to remove all our custom archive logic from Porter and and use archiver. 💯

PR #320, opened 2022-02-09
Add ArchiveAsync; context no longer optional

mholt:
Motivated by https://github.com/rclone/rclone/issues/2815.

Allows creating an archive by sending files down a channel rather than needing to pre-build a list of all the files up front. This is useful for efficiently creating an archive when reading a sequential stream. I think.

### DIFF
--- a/fs.go
+++ b/fs.go
@@ -221,6 +221,14 @@ type ArchiveFS struct {
 	Context context.Context // optional
 }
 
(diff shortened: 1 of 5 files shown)


mholt:
Merging, as the feedback from @ncw out-of-band is "That interface looks perfect and exactly what rclone needs" -- we can tweak it later if needed.

Issue #317, opened 2022-01-27
Brotli identification inconsistent


jhwz:
## What version of the package or command are you using?
v4.0.0-alpha.4

## What are you trying to do?
Open a `.br` file and read its contents through the Identify API 

## What steps did you take?

Here's a reproduction with a test file containing the given text that was compressed via a `brotli` CLI tool
```
func TestReproduce(t *testing.T) {
	const contents = "some text in a file\n"

	f, err := os.Open("testdata/compression/raw.br")
	require.NoError(t, err)

	format, err := archiver.Identify("testdata/compression/raw.br", f)
	require.NoError(t, err)

	rc, err := format.(archiver.Decompressor).OpenReader(f)
	require.NoError(t, err)

	data, err := io.ReadAll(rc)
	require.NoError(t, err)

	assert.Equal(t, contents, string(data))

	assert.NoError(t, rc.Close())
}
```

## What did you expect to happen, and what actually happened instead?
File should be consistently detected, instead sometimes it works OK and other times it returns one of the following errors:
```
matching rar: brotli: PADDING_1
matching zip: brotli: PADDING_1
```

## How do you think this should be fixed?
I think its returning inconsistent errors because the formats are stored in a map? Making the test non-deterministic. I'm not sure what the underlying issue is though!

## Bonus: What do you use archiver for, and do you find it useful?
We're using archiver for our CLI tooling to easily support a wide range of compression formats. Was trying out the new library version and caught this issue!

jhwz:
Ah interesting I'm finding a related issue with `.bz2` files sometimes: 

```matching zip: bzip2: corrupted input: invalid stream magic```

mholt:
That's very odd/interesting!...

Map iteration might be one factor, yes; we also do some stream seeking to be able to read from the beginning multiple times. (If that turns out to be problematic, another way to do that is to read the beginning of the stream and buffer it, then replace the input stream with a MultiReader that first reads from the buffer, then from the remaining stream... but that's more complicated.)

Brotli can only match by filename since brotli unfortunately doesn't have any particular header or preface/prefix to look for. Dunno if that matters, but thought I'd point that out.

Can you post a reproducible test case? Both the input file and a functional test (that just uses std lib). Then I can take a look at some point. Or if you want to dig in further, that'd be helpful too.

jhwz:
I'll add one soon but have a look first! If I can help and save you a bit of time then that would be ideal.

[...]

PR #322, opened 2022-02-19
Change Identify API to accept only an io.Reader

jhwz:
Implements a `"headerReader"` which caches all reads, and returns an `io.MultiReader` so we can return a reader which will read from the same point in the input stream. 

This is particularly useful for the `Identify` function, which can now just accept an `io.Reader` and not require an `io.Seeker` which is somewhat unnecessary. Also, it will also always return the Reader despite the error which the caller can then read from as normal.

On reflection it may be useful to still support `io.Seeker`'s in another function, maybe these changes should be moved somewhere else. It is more useful for files because you can still continue to use the `*os.File` instead of the more restrictive `io.Reader`.

Also includes some tests for the mentioned functions. 

This PR is based somewhat off the discussion in #317. I like this package and want to shift some of my projects that have a less clear API to using this. As an alternative, maybe if we expose the `formats` map, I can just implement this in a separate project.


mholt:
Thanks for this PR, I've been swamped catching up on things after a bout of sickness, but I like where this is going and I'll try to review it soon!

mholt:
@jhwz Thanks for this change. I took the liberty of refactoring a bit and renaming things to clean it up and make it more consistent with the existing code style. (I also moved the new code into formats.go and formats_test.go instead of their own separate files.) I also fixed a minor bug in the tests. I appreciate the thought that went into your change; I also think we don't have to do as much of the buffer management ourselves though, we can use existing types in the standard library, more like how I did it in Caddy's Layer 4 module: https://github.com/mholt/caddy-l4/blob/56bd7700d889f2ffd52353c241819ddbc7745ff6/layer4/connection.go#L72-L108

I think it'll be easier to maintain and is a little easier to read, and less error-prone.

Anyway, thank you for the contribution! I'm going to resolve the merge conflict and then finish this.

jhwz:
@mholt Just flicked through your changes - made a lot of sense. Thanks for merging - the new API will be much more convenient for my use cases!

mholt (approved):
Thanks again for the contribution. Let me know if you have any questions or feedback! :+1:

PR #332, opened 2022-04-14
Enhance FilesFromDisk (close #331)

### LINKED ISSUE #331: Using FilesFromDisk() for compressing files to root of archive
I wanna use archiver.FilesFromDisk for archiving files from folder to root of archive like this:
```
files, err := archiver.FilesFromDisk(nil, map[string]string{
	"/path/on/disk/folder/*":    "", // contents added recursively to root of archive without "folder"
})
```

I think it is useful feature.

Without this feature I need to write:
```
fileNames := make(map[string]string)
err := filepath.Walk(sourceDir, func(path string, info os.FileInfo, err error) error {
	if err != nil {
		return err
	}

	// skip root directory
	if path == sourceDir {
		return nil
	}

	fileNames[path] = ""
	return nil
})

if err != nil {
	return err
}

// map files on disk to their paths in the archive
files, err := archiver.FilesFromDisk(nil, fileNames)
```

mholt:
Now permit suffixing root on disk with separator to enumerate only
its contents and not to create that actual file in archive.

And attempt improving Windows compatibility. I don't have Windows handy so this'll have to be CI-based feedback.

@nicola-spb could you give this a try?

### DIFF
--- a/archiver.go
+++ b/archiver.go
@@ -49,6 +49,8 @@ func (f File) Stat() (fs.FileInfo, error) { return f.FileInfo, nil }
 // Map keys that specify directories on disk will be walked and added to the
 // archive recursively, rooted at the named directory. They should use the
 // platform's path separator (backslash on Windows; slash on everything else).
+// For convenience, map keys that end in a separator ('/', or '\' on Windows)
+// will enumerate contents only without adding the folder itself to the archive.
 //
 // Map values should typically use slash ('/') as the separator regardless of
 // the platform, as most archive formats standardize on that rune as the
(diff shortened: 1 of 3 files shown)


nicola-spb:
@mholt I am working in Linux too :(

mholt:
That's fine, I mostly want you to test the feature you requested.

nicola-spb:
Yes it's work on Ubuntu.

Tested it:
```
files, err := archiver.FilesFromDisk(nil, map[string]string{
	"/path/on/disk/folder/": "",  // adding files to root
})
```

```
files, err := archiver.FilesFromDisk(nil, map[string]string{
	"/path/on/disk/folder": "",  // adding files to "folder" in archive
})
```

mholt:
Perfect, thanks for verifying it does what you need!

I'll clean up the implementation and get them working on Windows. I might also work on the dot (`.`) situation but how complex it is depends whether it goes in this PR.

PR #353, opened 2022-10-14
zip: Fallback to in-memory buffer stream

johejo:
This might be preprocessing that the user of this package had better to do.

I have written this fallback several times when extracting remote zip files.

### DIFF
--- a/zip.go
+++ b/zip.go
@@ -177,15 +177,20 @@ func (z Zip) archiveOneFile(ctx context.Context, zw *zip.Writer, idx int, file F
 }
 
 // Extract extracts files from z, implementing the Extractor interface. Uniquely, however,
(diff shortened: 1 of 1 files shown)


mholt:
Thanks for the contribution. I think this would be fine, except _really big_ zip files, or many concurrent uses of this (like on a server) would quickly deplete memory, as there's no way to control the size of the buffer.

I wonder if the user should do this instead, to control memory? Or maybe we add a new field on the Zip struct specifying a maximum size for non-seeker inputs.

johejo:
Rather than adding options, I think it would be better to add documentation and control the users who read it.
Determining a default buffer size that satisfies many use cases is difficult.

Some users might prefer a limited buffer instead of io.ReadAll.
It may be better to divert golang/go's internal/safeio.
https://github.com/golang/go/blob/master/src/internal/saferio/io.go

https://github.com/golang/go/issues/47653

Issue #340, opened 2022-06-15
ArchiveFS.Open returns the first file in an implicit directory rather than a fs.ReadDirFile


jeremyje:
## What version of the package or command are you using?

v4.0.0-alpha.7

## What are you trying to do?

`ArchiveFS.Open()` with a zip file that does not have explicit directory entries.

## What steps did you take?

Adding a disabled test to showcase the bug via https://github.com/mholt/archiver/pull/339.

When you run the tests you'll get the following output.

```bash
$ go test ./...
--- FAIL: TestArchiveFS_ReadDir (0.00s)
    --- FAIL: TestArchiveFS_ReadDir/nodir.zip (0.00s)
        --- FAIL: TestArchiveFS_ReadDir/nodir.zip/Open(cmd) (0.00s)
            fs_test.go:136: 'cmd' did not return a fs.ReadDirFile, <nil>
        --- FAIL: TestArchiveFS_ReadDir/nodir.zip/Open(.github) (0.00s)
            fs_test.go:136: '.github' did not return a fs.ReadDirFile, <nil>
FAIL
FAIL    github.com/mholt/archiver/v4    0.064s
?       github.com/mholt/archiver/v4/cmd/arc    [no test files]
FAIL
```

Subtest of `TestArchiveFS_ReadDir` that reproduces this issue:

```go
// Uncomment to reproduce https://github.com/mholt/archiver/issues/340.
t.Run(fmt.Sprintf("Open(%s)", baseDir), func(t *testing.T) {
	f, err := fsys.Open(baseDir)
	if err != nil {
		t.Error(err)
	}

	rdf, ok := f.(fs.ReadDirFile)
	if !ok {
		t.Fatalf("'%s' did not return a fs.ReadDirFile, %+v", baseDir, rdf)
	}

	dis, err := rdf.ReadDir(-1)
	if err != nil {
		t.Fatal(err)
	}

	dirs := []string{}
	for _, di := range dis {
		dirs = append(dirs, di.Name())
	}

	// Stabilize the sort order
	sort.Strings(dirs)

	if diff := cmp.Diff(wantLS, dirs); diff != "" {
		t.Errorf("Open().ReadDir(-1) mismatch (-want +got):\n%s", diff)
	}
})
```
## What did you expect to happen, and what actually happened instead?

For ArchiveFS.Open(<directory>) will return the first file with that directory prefix.

## How do you think this should be fixed?

[...]

PR #354, opened 2022-10-15
rewrite file matching for fs Open, ReadDir and Stat

WeidiDeng:
Rewrite majority logic of fs Open, ReadDir and Stat
also fixes [340](https://github.com/mholt/archiver/issues/340)

ArchiveFS previously assume directory entries are always present in archive file listing when they are not. Also files under the same directory are stored together (add a test zip file that's a result of appending additional files to it).

Removed Top* utility func used to stat implicit directories.


mholt:
> Removed Top* utility func used to stat implicit directories.

I do use these in a project though.

I will review this soon! Kind of taking the weekend off :) Thanks

mholt:
@WeidiDeng Is there any specific reason the Top* utility functions were removed?

I'd be willing to consider this PR if we can keep those.

mholt:
@WeidiDeng Yeah, if we could add them back, I would appreciate it, since I use them in one of my projects :)

mholt (approved):
This is an impressive change. Thanks for working on it. I guess it's faster (and more correct)?

I really appreciate this! (merge at will)

Issue #350, opened 2022-10-07
Broken Zip archive


nicola-spb:
## What version of the package or command are you using?

v4.0.0-alpha.7

## What are you trying to do?

I trying extract ZIP archive which created by archiver.

## What steps did you take?

```

// "from" is directory with files
// "to" is path of archive
err = MakeArchive(context.Background(), from, to, archiver.CompressedArchive{
	Archival: archiver.Zip{},
})

// MakeArchive make archive of files in directory.
func MakeArchive(ctx context.Context, sourceDir string, archivePath string, format archiver.CompressedArchive) error {
	sep := string(filepath.Separator)

	// add "/" to the path "path/folder/"
	if !strings.HasSuffix(sourceDir, sep) {
		sourceDir = sourceDir + sep
	}

	// map files on disk to their paths in the archive
	files, err := archiver.FilesFromDisk(&archiver.FromDiskOptions{FollowSymlinks:  true}, map[string]string{
		sourceDir: "",
	})
	if err != nil {
		return err
	}

	// create the output file we'll write to
	out, err := os.Create(archivePath)
	if err != nil {
		return err
	}
	defer out.Close()

	// create the archive
	err = format.Archive(ctx, out, files)
	if err != nil {
		return err
	}

	return nil
}
```

## What did you expect to happen, and what actually happened instead?

When I creating archive and use archiver.FilesFromDisk (for all files in folder) with ZIP format then I have broken archive. I can open archive, open files in archive but I cannot extract files (in Ubuntu) and when I open archive in windows I see root folder but archive should not have it's folder.

## How do you think this should be fixed?

I think it's depends on archive type because when I create tar with gzip using MakeArchive it's ok.

## Please link to any related issues, pull requests, and/or discussion
See too https://github.com/mholt/archiver/issues/331

## Bonus: What do you use archiver for, and do you find it useful?
Using for reading and creating archive.

mholt:
Thanks, but this is pretty much the exact same info that's in your other post, so I don't know what to do with this. Please provide minimal reproduction steps from scratch so I can investigate, or, feel free to submit a PR with a fix. Thanks

nicola-spb:
Hi,

1) Create folder "test" ( "/home/.../test") and add some files.
2) Create folder "testresult" ( "/home/.../testresult")
3) Create main.go
`package main

import (
	"context"
	"fmt"
	"os"
	"path/filepath"
	"strings"

	"github.com/mholt/archiver/v4"
)

func main() {
[...]

PR #355, opened 2022-10-17
ignore empty root dir name when compressing file

WeidiDeng:
fix [350](https://github.com/mholt/archiver/issues/350)


mholt:
Thanks! I'll await confirmation from @nicola-spb and if it fixes the issue we can merge this in

WeidiDeng:
I find that FilesFromDisk will never return error, because it never checks error from filepath.Walk, (the error in walk func will stop walk but will not be returned).

mholt:
@WeidiDeng Wow, you are right. I missed that too :smile:

mholt (approved):
This is great, thank you very much for the patch @WeidiDeng, and thanks to @nicola-spb for confirming.

PR #364, opened 2023-01-07
add Archive method to SevenZip

### LINKED ISSUE #360: 7z should have Archive method
7z should implement `Archive` or other methods, which makes the `Identity` method unable to determine the file type

nikolaymatrosov:
close #360


mholt (approved):
Thank you!

PR #367, opened 2023-02-01
zip: fix File.Open not working if called after zip.Extract completes

ncw:
The zip.Extract method creates a closure for the File.Open function
pointer sent to the handler method.

Unfortunately it uses the same File for each call to the handler
function which means that if the caller stores the File objects and
uses them after Extract has returned then each File.Open function
pointer will open the same object.

This is easily fixed by making sure that each iteration of the Extract
loop uses a new File variable.

### DIFF
--- a/zip.go
+++ b/zip.go
@@ -202,6 +202,7 @@ func (z Zip) Extract(ctx context.Context, sourceArchive io.Reader, pathsInArchiv
 	skipDirs := skipList{}
 
 	for i, f := range zr.File {
+		f := f		// make a copy for the Open closure
 		if err := ctx.Err(); err != nil {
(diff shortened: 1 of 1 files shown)


ncw:
I found this bug when working on the rclone backend for the archiver: See https://github.com/rclone/rclone/issues/2815

Yes this is a lovely example of a classic Go gotcha!

mholt (approved):
Ah wow! I have started seeing static check / linter warnings for stuff like this in my editor recently. Thanks for the patch Nick!

mholt:
Memo to self: I don't think this is needed after Go 1.22 anymore: https://tip.golang.org/doc/go1.22#language

PR #369, opened 2023-02-03
Rework ArchiveAsync interface to make it return any archiving errors

### LINKED ISSUE #368: ArchiveAsync interface changes

I've been trying to use the `ArchiveAsync` interface to make an rclone backend.

It is good, but I'm having a problem with it: I need to know when the archiver has finished with the `File` I've passed in.

The problem is that at the time I create the `File` for the Archiver I have an open handle to pass to the archiver. However if the archiver go routine takes the handle after my function has returned, the open file handle will have been closed by layers above.

I think what is needed is instead of passing a `File` to the archiver, pass a struct like this

```go
type AsyncArchiveJob struct {
	File   File
	Result chan<- error
}
```

I think the archiver loops should then look something like this (here is the zip one for no particular reason)

```go
func (z Zip) ArchiveAsync(ctx context.Context, output io.Writer, jobs <-chan AsyncArchiveJob) error {
	zw := zip.NewWriter(output)
	defer zw.Close()

	var i int
	for job := range jobs {
		err := z.archiveOneFile(ctx, zw, i, job.File)
		job.Result <- err
		if err != nil {
			if z.ContinueOnError && ctx.Err() == nil { // context errors should always abort
				log.Printf("[ERROR] %v", err)
				continue
			}
			return err
		}
		i++
	}

	return nil
}
```

It may be that it doesn't actually need the continue on error at all here since we are now returning the errors to the user so this might be better:

```go
func (z Zip) ArchiveAsync(ctx context.Context, output io.Writer, jobs <-chan AsyncArchiveJob) error {
	zw := zip.NewWriter(output)
	defer zw.Close()

	var i int
	for job := range jobs {
		job.Result <- z.archiveOneFile(ctx, zw, i, job.File)
		i++
	}

	return nil
}
```

Thoughts?

If you think this is a good idea I'd be happy to make a PR.

This would make a backwards incompatible change :-(

We could
1. Do the change anyway and not update the /v4 since there are few/no users of ArchiveAsync and it is marked as being in alpha
2. Go to /v5
3. Work around by making a new `ArchiveAsyncError` interface (or something like that)

ncw:
This PR contains two commits

- Rework ArchiveAsync interface to make it return any archiving errors

After experience with the current ArchiveAsync interface while
creating an rclone backend it became clear that it wasn't quite right.
    
With the old interface, callers did not know when the File had been
archived and thus did not know when to release resources associated
with that file.
    
This patch changes the interface so that it returns the error
archiving each File. This means the caller can know when the File has
been archived and when to release resources. It returns the error for
each file archived which is useful too.
    
Fixes #368

- Add ArchiveAsync support to CompressedArchive

### DIFF
--- a/interfaces.go
+++ b/interfaces.go
@@ -60,6 +60,13 @@ type Archiver interface {
 	Archive(ctx context.Context, output io.Writer, files []File) error
 }
 
+// ArchiveAsyncJob contains a File to be archived and a channel that
+// the result of the archiving should be returned on.
+type ArchiveAsyncJob struct {
+	File   File
+	Result chan<- error
+}
+
 // ArchiverAsync is an Archiver that can also create archives
 // asynchronously by pumping files into a channel as they are
 // discovered.
@@ -67,9 +74,11 @@ type ArchiverAsync interface {
 	Archiver
 
 	// Use ArchiveAsync if you can't pre-assemble a list of all
-	// the files for the archive. Close the files channel after
+	// the files for the archive. Close the jobs channel after
 	// all the files have been sent.
-	ArchiveAsync(ctx context.Context, output io.Writer, files <-chan File) error
+	//
+	// This won't return until the channel is closed.
+	ArchiveAsync(ctx context.Context, output io.Writer, jobs <-chan ArchiveAsyncJob) error
 }
 
 // Extractor can extract files from an archive.
--- a/formats.go
+++ b/formats.go
@@ -235,6 +235,23 @@ func (caf CompressedArchive) Archive(ctx context.Context, output io.Writer, file
 	return caf.Archival.Archive(ctx, output, files)
 }
 
+// ArchiveAsync adds files to the output archive while compressing the result asynchronously.
+func (caf CompressedArchive) ArchiveAsync(ctx context.Context, output io.Writer, jobs <-chan ArchiveAsyncJob) error {
+	do, ok := caf.Archival.(ArchiverAsync)
+	if !ok {
+		return fmt.Errorf("%s archive does not support async writing", caf.Name())
+	}
+	if caf.Compression != nil {
+		wc, err := caf.Compression.OpenWriter(output)
+		if err != nil {
+			return err
+		}
+		defer wc.Close()
+		output = wc
+	}
+	return do.ArchiveAsync(ctx, output, jobs)
+}
+
 // Extract reads files out of an archive while decompressing the results.
 func (caf CompressedArchive) Extract(ctx context.Context, sourceArchive io.Reader, pathsInArchive []string, handleFile FileHandler) error {
 	if caf.Compression != nil {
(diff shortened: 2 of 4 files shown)


Joeynchas (approved):

ncw:
Having had a chance to sleep on it, I think the the interface I've proposed has a problem.

Allowing the range loop to exit on errors will cause a potential deadlock as any senders into `jobs` don't know whether there is a process listening or not. This isn't easily fixable in a race free way. So I think the loops should not exit on error.

So I think something like this is more correct.

```go
func (z Zip) ArchiveAsync(ctx context.Context, output io.Writer, jobs <-chan ArchiveAsyncJob) error {
	zw := zip.NewWriter(output)
	defer zw.Close()

	var i int
	for job := range jobs {
		job.Result <- z.archiveOneFile(ctx, zw, i, job.File)
		i++
	}

	return nil
}
```

I'll update the PR accordingly in a moment!

ncw:
I've added a new commit on top with that idea which I can squash into the first commit if you want.

interfaces.go
```diff
@@ -69,7 +76,9 @@ type ArchiverAsync interface {
 	// Use ArchiveAsync if you can't pre-assemble a list of all
 	// the files for the archive. Close the files channel after
 	// all the files have been sent.
-	ArchiveAsync(ctx context.Context, output io.Writer, files <-chan File) error
+	//
+	// This won't return until the channel is closed.
+	ArchiveAsync(ctx context.Context, output io.Writer, files <-chan ArchiveAsyncJob) error
```
mholt, line 81:
```suggestion
	ArchiveAsync(ctx context.Context, output io.Writer, jobs <-chan ArchiveAsyncJob) error
```

Wondering if maybe we should just keep this consistent with implementations.

mholt (approved):
Alrighty -- sorry for the wait, I was traveling with family.

I think this looks pretty good! The API is much better; and I guess context cancellation and ContinueOnError could be honored by the caller of `ArchiveAsync`.

My only comment is a nit and is optional. This has my approval. Thanks again!

interfaces.go
ncw: Ah yes, well spotted. Have fixed that now :-)

ncw:
I have sent a new update with the fix from above and I've also squashed the PR back into two commits with the error handling changes.

> Alrighty -- sorry for the wait, I was traveling with family.

No problem. You are doing much better than me at reviewing PRs :-)

> I think this looks pretty good! The API is much better; and I guess context cancellation and ContinueOnError could be honored by the caller of `ArchiveAsync`.

Yes its completely under the control of the caller. The current loops don't wait for context cancellation at the moment and I think if you need that it is better for the caller to do it otherwise you'll risk deadlocks when the input channel fills up because there is no-one listening.

> My only comment is a nit and is optional. This has my approval. Thanks again!

:+1:

mholt (approved):

PR #370, opened 2023-02-03
Allow Identify to take a nil io.Reader

ncw:
This PR contains two commits

- Allow Identify to take a nil io.Reader

Allow Identify to take a nil io.Reader
    
The Match functions can all take a nil io.Reader to mean just identify
the archive from its file name, but before this patch Identify could
not.
    
This patch allows the stream to be nil, and enables Identify to be
used to create formats for archives which haven't been created yet
just from their file name.

- Fix crash when Tar.Match is given a nil stream

### DIFF
--- a/formats.go
+++ b/formats.go
@@ -27,11 +27,14 @@ func RegisterFormat(format Format) {
 //
 // If no matching formats were found, special error ErrNoMatch is returned.
 //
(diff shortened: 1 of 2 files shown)


mholt (approved):
Interesting use case I hadn't thought of. Thanks for another great patch! :smiley:

Issue #358, opened 2022-11-28
archiver.FileSystem() alternative that takes io.Reader


tenox7:
## What would you like to have changed?
In addition to `archiver.FileSystem("filename.ext")` there should be a function that takes `io.Reader` or better `io.ReaderAt` and `size`. For example `archiver.NewFsReader(r io.ReaderAt, size int64)`.

## Why is this feature a useful, necessary, and/or important addition to this project?
This is necessary to use a file thats already open or for files that do not resize on OS filesystem and cannot be opened with `os.Open()`. For example [afero](https://github.com/spf13/afero) or any other virtual/in memory file system.

## What alternatives are there, or what are you doing in the meantime to work around the lack of this feature?
There are no alternatives, currently I can't use archiver.FileSystem with a file on a synthetic FS because archiver assumes the file can be opened with os.Open().

## Please link to any relevant issues, pull requests, or other discussions.
Go archive/zip implements NewReader in addition to file Open.

https://pkg.go.dev/archive/zip#NewReader

Similarly SevenZip:
https://pkg.go.dev/github.com/bodgit/sevenzip?utm_source=godoc#NewReader

Thank you!

mholt:
You want [`ArchiveFS`](https://pkg.go.dev/github.com/mholt/archiver/v4#ArchiveFS) -- simply specify the `Stream` field instead of the `Path` field.

Thanks for using archiver!

tenox7:
It worked, thank you! However it was pretty hard to figure out how to use this correctly. It was crashing on me until `Format` was specified from `Identify()` with a cast to `archiver.Archival`. Identify returns `io.Reader` to be used in further calls but I don't see an easy way to pass it directly to `ArchiveFS` which requires SectionReader. For this I used `io.NewSectionReader` from the opened file. But overall its little confusing. It could use an example. Thank you!

mholt:
Oh... if you don't know the format ahead of time, I guess that can be a little tricky; because Identify() reads some of the bytes, so you have to use the new reader it returns. I didn't realize you had an unknown format.

Maybe we can add an exported API for calling FileSystem on a stream...

mholt:
M'kay, I'll leave this open and give it some thought. I'd also accept a proposal to review for quicker turnaround!

aol-nnov:
Well, this is fantastic, thank you! But I have a more sophisticated case with which I'm, ahem, a bit stuck..

[...]

PR #379, opened 2023-04-29
Change Stream type to io.Reader

aol-nnov:
Wild guesses and musings on #358

@mholt please find my changes concerned the `ArchiveFS` `Stream` type. 

At first sight, it should not do any harm as zip archiver has type assertion inside and other types do not need a `SectionReader` (and your zip tests still pass :) )

### DIFF
--- a/fs.go
+++ b/fs.go
@@ -244,8 +244,8 @@ func (cf compressedFile) Close() error {
 // directory tree order.
 type ArchiveFS struct {
 	// set one of these
-	Path   string            // path to the archive file on disk, or...
-	Stream *io.SectionReader // ...stream from which to read archive
+	Path   string    // path to the archive file on disk, or...
+	Stream io.Reader // ...stream from which to read archive
 
 	Format  Archival        // the archive format
 	Prefix  string          // optional subdirectory in which to root the fs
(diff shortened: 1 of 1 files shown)


mholt:
Thanks for this!

I think this LGTM, but I want @tgulacsi to double-check since I know he was using the SectionReader as of #307, and although I imagine this change should be fine, I want to see what he says first :+1:

tgulacsi:
AFAIU this will advance the underlying Reader on Stat and ReadDir. I don't know whether this will cause problems with a .tar gz or not, just noticed it.

mholt:
@tgulacsi Thanks, although, I'm not sure I understand why:

> this will advance the underlying Reader on Stat and ReadDir.

If the value of the `Stream` field (type `io.Reader`) is actually a `*SectionReader`, wouldn't the behavior remain the same? Why would anything be different?

tgulacsi:
Yes, those changes use the Reader, which will advance if Read is called on them. (Not, if a new asectionReader is created.)

mholt:
Understood. Ok, that makes sense.

@aol-nnov We may need to preserve the SectionReader. I believe calling Read on the underlying reader otherwise is significant.

aol-nnov:
Okay, @mholt, limitations understood.

Issue #383, opened 2023-08-29
Tar files created inside the directory they are compressing are not walkable


drewstinnett:
## What version of the package or command are you using?
v4.0.0-alpha.8

## What are you trying to do?
I am attempting to read the table of contents information out of various compressed archives (similar to using `tar -tvf file.tar`). This is currently not working, if the tar was created using the current directory as it's first entry. For example:

```bash
❯ tar cvf test.tar .
a .
a ./test.tartar: ./test.tar: Can't add archive to itself

a ./1mb
a ./2mb
```

Even though it was tarred up to the current directory, and barfed out the warning `Can't add archive to itself`, I can still extract and get content listing using standard tar:

```bash
❯ tar tvf test-tar/test.tar
drwxr-xr-x  0 drews  staff       0 Aug 29 13:38 ./
-rw-r--r--  0 drews  staff 1048576 Aug 29 13:37 ./1mb
-rw-r--r--  0 drews  staff 2097152 Aug 29 13:37 ./2mb
❯ tar xvf test-tar/test.tar
x ./
x ./1mb
x ./2mb
```

However when I use the archiver.FileSystem type, it appears to get recursively hung on that first "./" entry, and it never reaches the next entry.

## What steps did you take?
Below is the sample code I am testing:

```go
package main

import (
	"context"
	"fmt"
	"io/fs"
	"log"
	"os"

	"github.com/mholt/archiver/v4"
)

func main() {
	if len(os.Args) != 2 {
		log.Fatal("must give an argument")
	}

	fsys, err := archiver.FileSystem(context.Background(), os.Args[1])
	if err != nil {
		log.Fatal(err)
	}
	err = fs.WalkDir(fsys, ".", func(path string, d fs.DirEntry, err error) error {
		fmt.Fprintf(os.Stderr, "Path: %v File: %v Dir: %v\n", path, d.Name(), d.IsDir())
		return nil
	})
	if err != nil {
		log.Fatal(err)
	}
}
```

## What did you expect to happen, and what actually happened instead?
I expected an output like this:

```bash
❯ go run ./ ~/Desktop/good.tar
Path: . File: good.tar Dir: true
Path: ._test-tar File: ._test-tar Dir: false
Path: test-tar File: test-tar Dir: true
Path: test-tar/._1mb File: ._1mb Dir: false
Path: test-tar/._2mb File: ._2mb Dir: false
Path: test-tar/1mb File: 1mb Dir: false
[...]

PR #384, opened 2023-08-30
Bugfix: Tar files created in their target directory shouldn't recurse infinitely

drewstinnett:
This addresses #383 .

Happy to refactor however you like, and open to any feedback. Thank you!


mholt (changes requested):
Thanks for the patch! I have not encountered this situation so I appreciate the test case. Is the right behavior really to just skip the file?

drewstinnett:
> Thanks for the patch! I have not encountered this situation so I appreciate the test case. Is the right behavior really to just skip the file?

Yep! At least to the best of my knowledge skipping this sub '.' entry keeps everything going, correctly. I tested it on some random archives here and they all behaved correctly with that weird entry being completely skipped.

fs.go
```diff
@@ -567,11 +567,15 @@ func (f ArchiveFS) ReadDir(name string) ([]fs.DirEntry, error) {
...
 		file.NameInArchive = strings.Trim(file.NameInArchive, "/")
+		if file.NameInArchive == "." {
+			return nil
+		}
 		files = append(files, file)
 		if file.NameInArchive == name && !file.IsDir() {
 			foundFile = true
 			return errStopWalk
 		}
+		}
```
mholt, line 578:
```suggestion
```

Oops, my bad

mholt (approved):
LGTM after I fix my syntax error

PR #385, opened 2023-09-03
Option to use number for user and/or group names

breezerider:
* add NumericUid and NumericGid to Tar struct
* writeFileToArchive: set Uname and Gname to empty string if NumericUid and NumericGid are true, respectively.

### DIFF
--- a/tar.go
+++ b/tar.go
@@ -17,11 +17,8 @@ func init() {
 }
 
 type Tar struct {
-	// If true, leave only numeric user id
-	NumericUid bool
-
-	// If true, leave only numeric group id
-	NumericGid bool
+	// If true, leave only numeric user and group id
+	NumericUidGid bool
 
 	// If true, errors encountered during reading or writing
 	// a file within an archive will be logged and the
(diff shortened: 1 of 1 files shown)


breezerider:
No real use case on my side, just left it like that for flexibility. Primary use case is to prevent aliasing and this means both flags are set simultaneously. Do you think it makes sense to join them?

mholt:
I think we should probably join them unless there's a valid use case for having them separated. We're still in alpha so we could change this later if there's a valid use case brought forth.

tar.go
```diff
@@ -17,6 +17,9 @@ func init() {
 }
 
 type Tar struct {
+	// If true, leave only numeric user and group id
+	NumericUidGid bool
```
mholt, line 21:
```suggestion
	// If true, preserve only numeric user and group id
	NumericUIDGID bool
```

(Just a nit, don't mind me. I'll still merge this in a moment.)

tar.go
```diff
@@ -81,6 +84,10 @@ func (Tar) writeFileToArchive(ctx context.Context, tw *tar.Writer, file File) er
 		return fmt.Errorf("file %s: creating header: %w", file.NameInArchive, err)
 	}
 	hdr.Name = file.NameInArchive // complete path, since FileInfoHeader() only has base name
+	if t.NumericUidGid {
```
mholt, line 87:
```suggestion
	if t.NumericUIDGID {
```

mholt (approved):
Thanks! This is good. I'm just changing `UidGid` to `UIDGID`, as dumb as that looks, because the convention I'm used to capitalizes initialisms :upside_down_face:

I appreciate the change!!

Issue #387, opened 2023-09-20
file 2G.file: writing data: archive/tar: write too long


halfcrazy:
## What version of the package or command are you using?

github.com/mholt/archiver/v4 v4.0.0-alpha.8

## What are you trying to do?

create an archive with continuous writing contents to the source file.

## What steps did you take?

prepare data with `dd if=/dev/urandom of=/2G.file bs=1M count=2048`

open a new window, and run scripts
```
while true; do
    echo "data..." >> 2G.file
    sleep 0.1
done
```

```
package main

import (
	"context"
	"github.com/mholt/archiver/v4"
	"os"
)

func main() {

	files, err := archiver.FilesFromDisk(nil, map[string]string{
		"/2G.file": "",
	})
	if err != nil {
		panic(err)
	}

	// create the output file we'll write to
	tmpFileName := "/tmp/aa"
	out, err := os.Create(tmpFileName)
	if err != nil {
		panic(err)
	}
	defer out.Close()

	// we can use the CompressedArchive type to gzip a tarball
	// (compression is not required; you could use Tar directly)
	format := archiver.CompressedArchive{
		Compression: archiver.Gz{},
		Archival:    archiver.Tar{},
	}

	// create the archive
	err = format.Archive(context.Background(), out, files)
	if err != nil {
		panic(err)
	}
}
```

## What did you expect to happen, and what actually happened instead?

Expect: no errors report
Actually: panic: file 2G.file: writing data: archive/tar: write too long

## Please link to any related issues, pull requests, and/or discussion

https://github.com/mholt/archiver/pull/8

## Bonus: What do you use archiver for, and do you find it useful?

My use case is to create a backup file for my living database.

mholt:
Thanks for the PR! I left a comment since I need help understanding it better.

PR #388, opened 2023-09-20
Fix create archive to a continuous writing source file failed

halfcrazy:
Fix https://github.com/mholt/archiver/issues/387

### DIFF
--- a/archiver.go
+++ b/archiver.go
@@ -221,8 +221,11 @@ func openAndCopyFile(file File, w io.Writer) error {
 		return err
 	}
 	defer fileReader.Close()
-	_, err = io.Copy(w, fileReader)
-	return err
+	_, err = io.CopyN(w, fileReader, file.Size())
+	if err != nil && err != io.EOF {
+		return err
+	}
+	return nil
 }
 
 // fileIsIncluded returns true if filename is included according to
--- a/archiver.go
+++ b/archiver.go
@@ -221,6 +221,9 @@ func openAndCopyFile(file File, w io.Writer) error {
 		return err
 	}
 	defer fileReader.Close()
+	// When file is in use and size is being written to, creating the compressed
+	// file will fail with "archive/tar: write too long." Using CopyN gracefully
+	// handles this.
 	_, err = io.CopyN(w, fileReader, file.Size())
 	if err != nil && err != io.EOF {
 		return err


halfcrazy:
I suppose the problem is caused by writing a header with the original file size and the reader reading the new append content.

image
0880c4c80710ddd68215
0880c4c80710ddd6a214



https://github.com/mholt/archiver/blob/v4.0.0-alpha.8/tar.go#L74-L100
```
func (Tar) writeFileToArchive(ctx context.Context, tw *tar.Writer, file File) error {
	if err := ctx.Err(); err != nil {
		return err // honor context cancellation
	}

	hdr, err := tar.FileInfoHeader(file, file.LinkTarget)
	if err != nil {
		return fmt.Errorf("file %s: creating header: %w", file.NameInArchive, err)
	}
	hdr.Name = file.NameInArchive // complete path, since FileInfoHeader() only has base name

==>	if err := tw.WriteHeader(hdr); err != nil {
		return fmt.Errorf("file %s: writing header: %w", file.NameInArchive, err)
	}

	// only proceed to write a file body if there is actually a body
	// (for example, directories and links don't have a body)
	if hdr.Typeflag != tar.TypeReg {
		return nil
	}

	if err := openAndCopyFile(file, tw); err != nil {
		return fmt.Errorf("file %s: writing data: %w", file.NameInArchive, err)
	}

	return nil
}
```

halfcrazy:
ping @mholt

mholt:
Thanks, sorry for the delay. Just had a baby. I just need to think about this one more time but I suppose it is likely to be merged soon 🙂

archiver.go
```diff
@@ -221,8 +221,11 @@ func openAndCopyFile(file File, w io.Writer) error {
 		return err
 	}
 	defer fileReader.Close()
-	_, err = io.Copy(w, fileReader)
-	return err
+	_, err = io.CopyN(w, fileReader, file.Size())
```
mholt, line 224: Can we add a comment here explaining why we're doing this instead of just io.Copy()? And why we ignore EOF errors below.

mholt (changes requested):
Ok, I mean, I think I have no problems with this, but I also don't really know what the implications are for _not_ your use case (which I don't fully understand still from code screenshots).

Could we add a comment maybe explaining why we're doing it this way?

Even better, tests to ensure correct behavior in your use case and the 'standard' use cases. I haven't really gotten around to doing a lot of testing in this package yet but if you want to make me even more confident in this change, then that'd be good.

archiver.go
halfcrazy: https://github.com/mholt/archiver/pull/8#issue-169892941

> When file is in use and size getting increased , create compressed file will failed and error is "copying contents: archive/tar: write too long". Changing to use CopyN to copy only the file size at that time, also checks io.EOF if file becomes small.

mholt:
```suggestion
	// When file is in use and size is being written to, creating the compressed
	// file will fail with "archive/tar: write too long." Using CopyN gracefully
	// handles this.
	_, err = io.CopyN(w, fileReader, file.Size())
```

mholt (approved):
Thanks for the improvement, we'll try it out

PR #394, opened 2024-01-22
7z: Fix iteration bug

SheltonZhu:
7z: Fix iteration bug that was already fixed in go 1.22.
The bug is same as https://github.com/mholt/archiver/blob/master/zip.go#L199

### DIFF
--- a/7z.go
+++ b/7z.go
@@ -82,6 +82,7 @@ func (z SevenZip) Extract(ctx context.Context, sourceArchive io.Reader, pathsInA
 	skipDirs := skipList{}
 
 	for i, f := range zr.File {
+		f := f // make a copy for the Open closure
 		if err := ctx.Err(); err != nil {
 			return err // honor context cancellation
 		}


mholt:
Thanks! Go 1.22 hasn't been released yet, and I doubt everyone will upgrade right away. I'll probably hold this open for a while until the release has been out for a while.

SheltonZhu:
> Thanks! Go 1.22 hasn't been released yet, and I doubt everyone will upgrade right away. I'll probably hold this open for a while until the release has been out for a while.

I mean I fix the "for" loops bug. 
Please merge my pr.
And the issue is same as zip.go
I copy code from  https://github.com/mholt/archiver/blob/master/zip.go#L199



ref: [*go-reference-in-for-loop-sets-the-wrong-reference*](https://stackoverflow.com/questions/62929600/go-reference-in-for-loop-sets-the-wrong-reference)

SheltonZhu:
And go 1.22 has fix it. (No more sharing of loop variables)
ref: https://antonz.org/go-1-22/

SheltonZhu:
I can't get the excepted `io.ReadCloser` when I call `archiver.File.Open()` outside of the `archiver.FileHandler`.
But it is correct when I use `archiver.Zip{}`. (Translated from GPT)
```go
type ArchiverExtractor struct {
	archiver.Extractor
	sourceArchive   io.Reader
	fileHandlerFunc FileHanderFunc
	pathsInArchive  []string
}

func (ae *ArchiverExtractor) ExtractFile(ctx context.Context, filePath string) (*archiver.File, error) {
	files := make([]archiver.File, 0)
	ff := FileFilter(&files, filePath)
	if ae.fileHandlerFunc != nil {
		ff = ae.fileHandlerFunc(&files)
	}
	err := ae.Extract(ctx, ae.sourceArchive, ae.pathsInArchive, ff)
	if len(files) == 0 {
		return nil, fmt.Errorf("file not found")
	}
	return &files[0], err
}

// ...
func main() {
    // ...

    frc, err := f.Open() 
    if err != nil {
       panic(err)
    }
    defer frc.Close()
    fileBytes, _ := io.ReadAll(frc) // empty 
}
```

mholt:
Oh, my bad -- I understand now. Thank you.

mholt (approved):

PR #401, opened 2024-04-06
Add lzip support

### LINKED ISSUE #400: Add lzip support
## Introduce the format you are requesting.

lzip is a compressed file format that using LZMA developed in 2008. The compression ratio is about the same as xz (which uses LZMA2). It mainly features a simpler file format and therefore easier error recovery. GMP also distributes this format of tarballs in addition to xz and zstd.[^1]

## What do YOU use this format for?

Use cases of this format are similar to gzip and bzip2.

## What is the format's conventional file extension(s)?

```text
.lz
```

Please see <https://datatracker.ietf.org/doc/html/draft-diaz-lzip-09#section-4.1>.

## Please link to the format's formal or official specification(s).

- <https://www.nongnu.org/lzip/manual/lzip_manual.html#File-format>
- <https://datatracker.ietf.org/doc/html/draft-diaz-lzip-09#section-2>

## Which Go libraries could be used to implement this format?

- <https://pkg.go.dev/github.com/sorairolake/lzip-go>

[^1]: https://gmplib.org/#DOWNLOAD

sorairolake:
Closes #400

### DIFF
--- a/lzip.go
+++ b/lzip.go
@@ -3,6 +3,7 @@ package archiver
 import (
 	"bytes"
 	"io"
+	"path/filepath"
 	"strings"
 
 	"github.com/sorairolake/lzip-go"
@@ -21,7 +22,7 @@ func (lz Lzip) Match(filename string, stream io.Reader) (MatchResult, error) {
 	var mr MatchResult
 
 	// match filename
-	if strings.Contains(strings.ToLower(filename), lz.Name()) {
+	if filepath.Ext(strings.ToLower(filename)) == lz.Name() {
 		mr.ByName = true
 	}
 
--- a/lzip.go
+++ b/lzip.go
@@ -1,0 +1,52 @@
+package archiver
+
+import (
+	"bytes"
+	"io"
+	"strings"
+
+	"github.com/sorairolake/lzip-go"
+)
+
+func init() {
+	RegisterFormat(Lzip{})
+}
+
+// Lzip facilitates lzip compression.
+type Lzip struct{}
+
+func (Lzip) Name() string { return ".lz" }
+
+func (lz Lzip) Match(filename string, stream io.Reader) (MatchResult, error) {
+	var mr MatchResult
+
+	// match filename
+	if strings.Contains(strings.ToLower(filename), lz.Name()) {
+		mr.ByName = true
+	}
+
+	// match file header
+	buf, err := readAtMost(stream, len(lzipHeader))
+	if err != nil {
+		return mr, err
+	}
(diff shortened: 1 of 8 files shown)


mholt:
Nice, thank you! This is looking good. Will merge this shortly. Appreciate your contribution!

mholt:
Hm, lz4 compression with extension is failing. @sorairolake Does it fail for you locally?

mholt:
Hmm, sounds like something funky with the test suite. I'll look into it, but your change seems sound.

sorairolake:
Both lz4 and lzip starts with "lz". So I think this may be related.

lzip.go
```diff
@@ -0,0 +1,52 @@
...
+// Lzip facilitates lzip compression.
+type Lzip struct{}
+
+func (Lzip) Name() string { return ".lz" }
+
+func (lz Lzip) Match(filename string, stream io.Reader) (MatchResult, error) {
+	var mr MatchResult
+
+	// match filename
+	if strings.Contains(strings.ToLower(filename), lz.Name()) {
```
mholt, line 24: Ah, indeed -- make this comparison more strict and that should hopefully fix the test.

sorairolake: I think this can be fixed with the following changes:

```suggestion
	if filepath.Ext(strings.ToLower(filename)) == lz.Name() {
```

However, this requires that the extension be the last element of the path.

mholt: That would fail for `.lzip.gz` as you say, but maybe a check for both a suffix of `.lz` or the string contains `.lz.`?

sorairolake: I think there is almost no possibility that `.lz` and `.lz4` exists anywhere other than at the end of the path.

lz4.go
```diff
@@ -23,7 +24,7 @@ func (lz Lz4) Match(filename string, stream io.Reader) (MatchResult, error) {
 	var mr MatchResult
 
 	// match filename
-	if strings.Contains(strings.ToLower(filename), lz.Name()) {
+	if filepath.Ext(strings.ToLower(filename)) == lz.Name() {
```
mholt, line 27: Why was this one changed?

lzip.go
mholt: That's a good point.

lz4.go
sorairolake: I thought it would be a conflict, but it was a misunderstanding.

mholt (approved):
Awesome, looks good now :) Thank you!

mholt:
PS. I hope to travel to Japan someday :D

PR #402, opened 2024-05-20
Error messages: add new lines

martinetd:
When extracting a large archive with many ignored errors through syft the errors do not go to the next line which is hard to read.

Add trailing new line.

(cc @wagoodman for heads up as this is mostly for syft -- assuming it'll get in eventually after merging here anyway but might as well be aware)

### DIFF
--- a/7z.go
+++ b/7z.go
@@ -111,7 +111,7 @@ func (z SevenZip) Extract(ctx context.Context, sourceArchive io.Reader, pathsInA
 			skipDirs.add(dirPath)
 		} else if err != nil {
 			if z.ContinueOnError {
-				log.Printf("[ERROR] %s: %v", f.Name, err)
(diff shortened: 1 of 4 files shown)


martinetd:
(By the way some of these "ERROR" messages aren't actually errors as far as syft is concerned (only skipped files/symlinks) so it'd actually be nice to have a way not to print them at all (perhaps as a debug toggle?), but for now just new lines would be a great improvement)

martinetd:
Hm, sorry go's Errorf does add a new line doesn't it.
It looks like the logs just interact badly with syft's "pretty" output, as running in quiet mode lists errors correctly.

Closing this, sorry for the noise. If I find time I might eventually open something else to be optionally less verbose about some type of errors, but that'll have to wait.

Issue #405, opened 2024-06-07
DMG/`bzip2` file read error: `bzip2: corrupted input: invalid stream magic`


rgmz:
## What version of the package or command are you using?
v4.0.0-alpha.8

## What are you trying to do?
Decompress a `.dmg` file detected as `.bzip2`.

```sh
$ file -i BitcoinTemplate.dmg
BitcoinTemplate.dmg: application/x-bzip2; charset=binary
```

Note that there is a warning about "trailing garbage" when decompressing it using the CLI, however, it still succeeds.
```sh
$ bzip2 -d BitcoinTemplate.dmg
bzip2: Can't guess original name for BitcoinTemplate.dmg -- using BitcoinTemplate.dmg.out

bzip2: BitcoinTemplate.dmg: trailing garbage after EOF ignored
```

## What steps did you take?

1. Download https://github.com/cicorias/bitcoin/raw/a4f2c8419f5a8b47c4b48e962b4ff02885164c9b/contrib/BitcoinTemplate.dmg
2. Run the following test
    ```go
    func TestDmg(t *testing.T) {
	    f, err := os.Open("/tmp/BitcoinTemplate.dmg")
	    require.NoError(t, err)
    
	    format, _, err := archiver.Identify("/tmp/BitcoinTemplate.dmg", f)
	    require.NoError(t, err)
    
	    rc, err := format.(archiver.Decompressor).OpenReader(f)
	    require.NoError(t, err)
    
	    _, err = io.ReadAll(rc)
	    require.NoError(t, err)
    
	    assert.NoError(t, rc.Close())
    }
    ```
3. Receive error from `io.ReadAll`
    ```
     Error:      	Received unexpected error:
        	         bzip2: corrupted input: invalid stream magic
    ```

## What did you expect to happen, and what actually happened instead?

I expected the file to be identified and decompressed successfully.

## How do you think this should be fixed?
Unsure.

## Please link to any related issues, pull requests, and/or discussion
https://github.com/trufflesecurity/trufflehog/issues/2935

mholt:
As the [docs](https://pkg.go.dev/github.com/mholt/archiver/v4#Identify) say:

> If stream is non-nil then the returned io.Reader will always be non-nil and will read from the same point as the reader which was passed in; it should be used in place of the input stream after calling Identify() because it preserves and re-reads the bytes that were already read during the identification process. 

You're passing in a stream that gets read, but ignoring the returned stream (`_`) causing corruption to the reader.

[...]

PR #407, opened 2024-06-13
Expose gzip.Reader.Multistream

rgmz:
Somewhat related to #405, it turns out that `compress/gzip` has a feature to handle trailing data without returning an error.

> If you would like the ability to read a valid gzip member and specially handle non-gzip data after the member, then you can use the [gzip.Reader.Multistream](https://pkg.go.dev/compress/gzip#Reader.Multistream) feature to read the file one member at a time.
> 
> https://github.com/golang/go/issues/47809

I created a field called `DisableMultistream` and not `Multistream` because `gzip.Reader.Multistream` is enabled by default. The latter would silently disable `Multistream` for any existing instances of `archiver.Gz{}`. 

**Additional Context**

https://github.com/trufflesecurity/trufflehog/issues/2934

### DIFF
--- a/gz.go
+++ b/gz.go
@@ -20,6 +20,10 @@ type Gz struct {
 	// than no compression.
 	CompressionLevel int
 
+	// DisableMultistream controls whether the reader supports multistream files.
+	// See https://pkg.go.dev/compress/gzip#example-Reader.Multistream
+	DisableMultistream bool
+
 	// Use a fast parallel Gzip implementation. This is only
 	// effective for large streams (about 1 MB or greater).
 	Multithreaded bool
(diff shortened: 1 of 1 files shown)


gz.go
```diff
@@ -65,14 +69,21 @@ func (gz Gz) OpenWriter(w io.Writer) (io.WriteCloser, error) {
...
+		}
+
+		return gzR, err
 	} else {
-		rc, err = gzip.NewReader(r)
+		gzR, err := gzip.NewReader(r)
+		if gzR != nil && gz.DisableMultistream {
+			gzR.Multistream(false)
+		}
+
```
mholt, line 84:
```suggestion
```

gz.go
```diff
@@ -65,14 +69,21 @@ func (gz Gz) OpenWriter(w io.Writer) (io.WriteCloser, error) {
...
 func (gz Gz) OpenReader(r io.Reader) (io.ReadCloser, error) {
-	var rc io.ReadCloser
-	var err error
 	if gz.Multithreaded {
-		rc, err = pgzip.NewReader(r)
+		gzR, err := pgzip.NewReader(r)
+		if gzR != nil && gz.DisableMultistream {
+			gzR.Multistream(false)
+		}
+
```
mholt, line 77:
```suggestion
```

gz.go
```diff
@@ -65,14 +69,21 @@ func (gz Gz) OpenWriter(w io.Writer) (io.WriteCloser, error) {
...
+		return gzR, err
 	} else {
```
mholt, line 79: Since the prior block ends in `return` we can drop the `else` block entirely.

mholt (changes requested):
Thanks! I have the stupidest style nits to pick with this :sweat_smile:  but then we can get it merged in right away.

rgmz:
Done!

mholt (approved):
Cool -- let's try it. Thanks!

PR #411, opened 2024-07-01
go.mod: Update STARRY-S/zip

franchb:
Commits https://github.com/STARRY-S/zip/compare/v0.1.0...v0.2.0


mholt:
Does this fix anything that is broken? I'm hesitant to upgrade dependencies these days unless there's significant improvements. We also need to avoid updating the go versions in go.mod since they can break ecosystems pretty quickly.

PR #419, opened 2024-08-21
zip: Fix compression method not set without SelectiveCompression enabled (closes #418)

### LINKED ISSUE #418: Zip{} does not compress files without SelectiveCompression enabled
## What version of the package or command are you using?

https://github.com/mholt/archiver/commit/743ede3881bb34d43b9ca063232367cbc237daea

## What are you trying to do?

I'm trying to make a zip archive with compression.

## What steps did you take?

```go
format := archiver.Zip{
	Compression:          zip.Deflate,
}
format.Archive(context.Background(), writer, file)
```

Expected: the files should be compressed.
Actual: the files are not compressed at all.

We should append the else-clause in zip.go (last 3 lines in the code snippet below):
```go
	if file.IsDir() {
		if !strings.HasSuffix(hdr.Name, "/") {
			hdr.Name += "/" // required
		}
		hdr.Method = zip.Store
	} else if z.SelectiveCompression {
		// only enable compression on compressable files
		ext := strings.ToLower(path.Ext(hdr.Name))
		if _, ok := compressedFormats[ext]; ok {
			hdr.Method = zip.Store
		} else {
			hdr.Method = z.Compression
		}
	} else {
	        hdr.Method = z.Compression
	}
```

## Workaround

It works as expected when I enable SelectiveCompression:
```go
format := archiver.Zip{
	SelectiveCompression: true,
	Compression:          zip.Deflate,
}
format.Archive(context.Background(), writer, file)
```

hevav:


mholt (approved):
This change makes sense. Thank you!

Issue #391, opened 2023-12-18
Unarchiving read-only folders fails


omerzi:
## What version of the package or command are you using?
v3.5.1

## What are you trying to do?
Trying to extract a tar.gz file (which I suspect might replicate with other archive formats) that contains a read-only folder along with its files.

## What steps did you take?
Create an archive with a write-protected folder:

1. mkdir -p TestArchive/writeprotectedfolder
2. touch TestArchive/writeprotectedfolder/file1.txt
3. touch TestArchive/writeprotectedfolder/file2.txt
4. chmod -w TestArchive/writeprotectedfolder/
5. tar -xzf example.tar.gz
6. Untar using the Unarchive function

## What did you expect to happen, and what actually happened instead?
An error returned:
`reading file in tar archive: /path/TestArchive/writeprotectedfolder/file2.txt: creating new file: open /path/TestArchive/writeprotectedfolder/file2.txt: permission denied`

## How do you think this should be fixed?
The issue arises when the code creates the `writeprotectedfolder` with its original read-only permissions (https://github.com/mholt/archiver/blob/v3.5.1/tar.go#L264). Consequently, we encounter an inability to write `file1.txt` and `file2.txt` into it. I intended to initiate a pull request to address this, but I am unable to do so as the master branch points to v4, and there isn't a v3 branch available.

## Please link to any related issues, pull requests, and/or discussion
I believe that this pull request has caused the issue: https://github.com/mholt/archiver/issues/133.

mholt:
I'm not working on v3 anymore. But this is good to know for v4, in case we do add unarchiving capabilities: https://github.com/mholt/archiver/issues/311

Maybe a better solution would be to apply folder permissions after extraction is complete.

PR #425, opened 2024-10-24
add unarchiver example

### LINKED ISSUE #311: Easier unarchiving support with v4?
Hello, thanks for the awesome library. The new version is fantastic since I can use it in conjunction with other network libraries to stream HTTP response right into unarchiving process.

I'm using archiver on Google Cloud Storage caching library for GitHub Actions, and while archiving files was [super easy](https://github.com/shakrmedia/gcs-cacher/blob/minku/include-symlink/cacher/cacher.go#L138), unarchiving was [more involved](https://github.com/shakrmedia/gcs-cacher/blob/minku/include-symlink/cacher/cacher.go#L252) with some manual file type handling.

Is there a plan to have a shorthand to make unarchival process a bit easier?

Quick way to unarchive file on v4, similar to how v3 works in some way

Currently, unarchiving requires a user to manually handle each different Tar type flags which involves a lot of boilerplate. 

I took a look at v3 code and [implemented unarchiving in-line](https://github.com/shakrmedia/gcs-cacher/blob/minku/include-symlink/cacher/cacher.go#L252), however it's not ideal at least from my perspective.

jm33-m0:
Hi Matt, thanks for the great work!

I noticed there's no `Unarchive` function in v4 yet, and the documentation is incomplete and hard to follow, so I wrote a small program that invokes your extractor API to automatically identify the format and unarchive tarballs.

Please check the code and point out where I need to improve. It can serve as an example for users who seek a quick solution from this project.


mholt:
Thanks for this; I like the idea of posting an example instead of prescribing one way to do it in the code itself.

I have a few suggestions on how to tweak this but I will have to come back and write those up later.

I suppose this might resolve #311.

mholt:
Does this example take into account #391?

jm33-m0:
> Does this example take into account #391?

In fact, my code doesn't retain the file/dir permissions. So I guess it will "solve" #391.

mholt:
Gotcha. This is a good start, but I think file mode bits (permissions, etc) are important, so we might need to include those in a proper example.

jm33-m0:
I have updated the code with necessary changes to address the directory permission issue, and I also added code to handle symlinks.

mholt:
Thank you! However, I don't think we should be providing an actual program, just an example in the docs.

I'm actually about the archive this repository, but we can continue this discussion over in [mholt/archives](https://github.com/mholt/archives) if you would like.

